PR #3, opened 2014-12-21
Add Migration Support

cah:


jhthorsen:
This is very tempting to accept. Before I dig any deeper, could you tell me how much this is alike Mojo::Pg::Migrations? Is it the same, except of the name?

There is a failing test though:
https://travis-ci.org/jhthorsen/mojo-mysql/jobs/44767456

cah:
Yeah, it's almost identical. Changed the name of the variables from pg to mysql.

jhthorsen:
I think we should mention how risky migrations are in MySQL, compared to Postgres. The lack of transactions and if something goes bad, there's no way to know where you end up...

As a disclaimer.

I will check if there's any more things MySQL lack in this context.

cah:
I'm working on MySQL support for Minion. As it stands, it requires migration support but I could remove it.

jhthorsen:
I will merge this one way or the other. Just not tonight.

You should probably add the same disclaimer to the minion backend, once I've managed to put something together. Like "remember to back up your database first".

Thank you.

cah:
I've added the additional documentation and  warning concerning mysql transactions. If you have additional recommendations, let me know.

jhthorsen:
Thank you! Will soon make a new release to CPAN.

lib/Mojo/mysql/Migrations.pm
```diff
@@ -0,0 +1,231 @@
...
+
+  # Unknown version
+  my ($up, $down) = @{$self->{migrations}}{qw(up down)};
+  croak "Version $target has no migration" if $target != 0 && !$up->{$target};
+
+  # Already the right version (make sure migrations table exists)
+  my $db = $self->mysql->db;
+  return $self if $self->_active($db) == $target;
+
+  # Lock migrations table and check version again
```
jhthorsen, line 49: So... what's actually going on here in your version @cah ? Looks like there's some code removed... Don't get what the rest of the lines from 49:53 does.

lib/Mojo/mysql.pm
```diff
@@ -149,6 +156,19 @@ Data Source Name, defaults to C<dbi:mysql:dbname=test>.
 Maximum number of idle database handles to cache for future use, defaults to
 C<5>.
 
+=head2 migrations
+
+MySQL does not support DDL transactions. B<Therefore, migrations should be used with extreme caution. Backup your database. You've been warned.>
```
jhthorsen, line 161: I think this warning should be added to the Mojo::mysql::Migration module as well.

lib/Mojo/mysql/Migrations.pm
```diff
@@ -0,0 +1,231 @@
...
+  -- 2 down
+  drop table stuff;
+
+The idea is to let you migrate from any version, to any version, up and down.
+Migrations are very safe, because they are performed in transactions and only
+one can be performed at a time. If a single statement fails, the whole
+migration will fail and get rolled back. Every set of migrations has a
+L</"name">, which is stored together with the currently active version in an
+automatically created table named C<mojo_migrations>.
+
```
jhthorsen, line 138: I think we should add the warning here. Something like this:

```
=head1 WARNING

MySQL does not support DDL transactions. Therefore, migrations should be used with extreme caution. Backup your database. You've been warned!
```

lib/Mojo/mysql/Migrations.pm
cah, line 49: The table lock on mojo_migrations was removed. I wasn't able to make it function like the Postgres module. 49-53 checks the current version and exits the migrations if the database is on the correct version.

jhthorsen, line 49: But the code says "Lock migrations table and check version again" which I find very confusing if the table lock was removed.

Also, why do you start a transaction if mysql doesn't care?

PR #6, opened 2015-01-13
Disable Statement cache

harry-bix:
Do not cache statements in Mojo::mysql::Database.
$sth->finish is implicitly called when $sth is destroyed, fixing problem with mysql_use_result
Synced changes from Mojo::Pg

### DIFF
--- a/lib/Mojo/mysql.pm
+++ b/lib/Mojo/mysql.pm
@@ -149,8 +149,8 @@ L<Mojo::mysql> inherits all events from L<Mojo::EventEmitter> and can emit the
 following new ones.
 
 =head2 connection
-  $pg->on(connection => sub {
-    my ($pg, $dbh) = @_;
+  $mysql->on(connection => sub {
+    my ($mysql, $dbh) = @_;
     ...
   });
 
@@ -200,7 +200,7 @@ C<mysql_auto_reconnect> is never enabled, L<Mojo::mysql> takes care of dead conn
 
 C<AutoCommit> cannot not be disabled, use $db->L<begin|Mojo::mysql::Database/begin> to manage transactions.
 
-C<RaiseError> is disabled event loop for asyncronous queries.
+C<RaiseError> is disabled in event loop for asyncronous queries.
 
 =head2 password
 
--- a/lib/Mojo/mysql.pm
+++ b/lib/Mojo/mysql.pm
@@ -15,7 +15,7 @@ has migrations      => sub {
   weaken $migrations->{mysql};
   return $migrations;
 };
-has options => sub { {AutoCommit => 1, PrintError => 0, RaiseError => 1} };
+has options => sub { {mysql_enable_utf8 => 1, AutoCommit => 1, PrintError => 0, RaiseError => 1} };
 has [qw(password username)] => '';
 
 our $VERSION = '0.04';
@@ -153,7 +153,7 @@ following new ones.
     my ($pg, $dbh) = @_;
     ...
   });
-  
+
 Emitted when a new database connection has been established.
 
 =head1 ATTRIBUTES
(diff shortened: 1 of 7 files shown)


lib/Mojo/mysql.pm
```diff
@@ -72,6 +72,10 @@ sub _dequeue {
   # you, silently, but only if certain env vars are set
   # hint: force-set mysql_auto_reconnect or whatever it's called to 0
   $dbh->{mysql_auto_reconnect} = 0;
+  # Maintain Commits with Mojo::mysql::Transaction
+  $dbh->{AutoCommit} = 1;
```
jhthorsen, line 76: Won't this override any AutoCommit setting in the options attribute?

lib/Mojo/mysql.pm
```diff
@@ -72,6 +72,10 @@ sub _dequeue {
...
+
+  $self->emit(connection => $dbh);
```
jhthorsen, line 78: Is this related to the solution, or is it just to be more Mojo::Pg compatible?

lib/Mojo/mysql/Database.pm
```diff
@@ -2,14 +2,12 @@ package Mojo::mysql::Database;
 use Mojo::Base 'Mojo::EventEmitter';
 
 use DBD::mysql;
-use IO::Handle;
```
jhthorsen, line 5: Nice catch :+1:

lib/Mojo/mysql/Database.pm
```diff
@@ -58,31 +58,13 @@ sub query {
...
-}
-
 sub _next {
   my $self = shift;
 
   return unless my $next = $self->{waiting}[0];
   return if $next->{sth};
 
-  my $sth = $next->{sth} = $self->_dequeue(1, $next->{query});
+  my $sth = $next->{sth} = $self->dbh->prepare($next->{query}, {async => 1});
```
jhthorsen, line 67: Won't this make all statements async..? Doesn't look like it from the test results though. (There's something I'm missing here)

jhthorsen:
I'm sorry for not being able to understand all of the code you have put in. I would like to make you co-auth of the module, since it seems like you have a better grasp of how this works than me.

Would you like to be co-auth?

Either way, I would very much appreciate if you could answer my questions, so I can understand it better.

lib/Mojo/mysql.pm
harry-bix, line 76: Yes, it will override it. But setting AutoCommit = 0, needs explicit $dbh->commit, or $dbh->disconnect. We have Mojo::mysql::Transaction to deal with transactions.

lib/Mojo/mysql.pm
harry-bix, line 78: Mojo::Pg does not have it.

lib/Mojo/mysql.pm
jhthorsen, line 76: Could you mention this in the documentation? Right now, the doc explicit mention AutoCommit: https://github.com/harry-bix/mojo-mysql/blob/no-sth-cache/lib/Mojo/mysql.pm#L194

lib/Mojo/mysql/Database.pm
harry-bix, line 67: Yes, _next is called only for non-blocking queries.

lib/Mojo/mysql.pm
jhthorsen, line 78: I was thinking about the "connection" event.

lib/Mojo/mysql/Database.pm
jhthorsen, line 67: Ah! Of course. Thank you :smile:

lib/Mojo/mysql.pm
harry-bix, line 78: Yes, it is just Mojo::Pg compatible.
I found it useful as better alternative to DBD::mysql mysql_init_command

jhthorsen, line 78: Nice :+1:

harry-bix:
Sure. I will be glad to be co-author.

jhthorsen:
Do you have a PAUSE/CPAN user id? I've added you as co-auth to this project now.

harry-bix:
PAUSE id is HARRYL

jhthorsen:
@harry-bix: What's the status on this pull requst?

lib/Mojo/mysql.pm
```diff
@@ -15,7 +15,7 @@ has migrations      => sub {
   weaken $migrations->{mysql};
   return $migrations;
 };
-has options => sub { {AutoCommit => 1, PrintError => 0, RaiseError => 1} };
+has options => sub { {mysql_enable_utf8 => 1, AutoCommit => 1, PrintError => 0, RaiseError => 1} };
```
jhthorsen, line 18: Doesn't this break backward compatibility?

harry-bix:
Applications that use character sets different than ascii or utf-8 could be affected by this change.
Normally these applications do not rely on the default negotiated character set and handle conversion explicitly.

jhthorsen:
I'm not sure if I'm a big fan, but I'm not going to revert it. I trust the change, but I would prefer if we don't change defaults without a good reason.

PR #9, opened 2015-03-02
Fully async implementation of MySQL protocol

harry-bix:
Async support in DBD::mysql is partial and buggy.
Mojo::mysql::Connection is fully async implementation of MySQL client protocol version 10. (MySQL versions after 4.1).
Mojo::mysql is rewritten to use Mojo::mysql::Connection instead of DBI and DBD::mysql, keeping as much compatibility as possible.
Most incompatible changes are:
- removing of $db->dbh, $db->dsn and $result->sth.
- some features of MySQL protocol are not implemented (unix socket, compression, TLS, server side prepared statements, older protocol version)


jhthorsen:
This is _not_ back compat at all.
- You're completely replacing DBD::mysql. I don't know if I want that. I want to be able to choose.
- You're adding a bunch of host, port, database attributes. Not sure why you need those. Just pass on a URL object to the connection.
- db() is (optionally) async. I don't see why connect() can't be lazy. (delayed to whenever you run a query)
- You should also be able to parse a dsn.
- Why can't you store your connection object in the dbh attribute? if the name is wrong: you could rename your module to Mojo::mysql::Handle.
- do('START TRANSACTION'); seems very stateless. I don't think you should rely on MySQL on keeping transaction state for you. Also, that do() seems blocking.
- _expand_sql and friends should probably be moved to a Mojo::mysql::Util module.

There could be more...

When that is said: I really like the initiative and I really want this to be part of the API, but for all of your changes to go through, you need:
- People telling me you're right (or I'm wrong, hehe)
- A more compatible API with Mojo::Pg

harry-bix:
I was looking for a module that _"makes MySQL a lot of fun to use with the Mojolicious real-time web framework"_.
Being _"a tiny wrapper around DBD::mysql"_ is just a way of achieving this.

Current version of Mojo::mysql takes the most of DBD::mysql, unfortunately in terms of async execution it is very limitted and has a lot of issues. It is based on undocumented mysqlclient C-API functions.

The idea of this fork is to replace use of DBD::mysql with the minimum needed functionality for making async connect and async query execution using TCP connection and well documented connection protocol.
Compatibility with Mojo::Pg and current version of Mojo::mysql is on focus, not compatibility with DBI/DBD.

In both Mojo::Pg and Mojo::mysql access to underlying handles dbh and sth is needed for things like:
**last_insert_id**, **affected_rows**, **more_results**, column descriptions, detailed error information.
Isn't it better to pack what is needed in ::Database and ::Results methods instead of relying on DBI calls, which are in many ways different between Pg and mysql.

What I am proposing is to release a new module of this fork.
Add aditional methods in ::Database and ::Results in both modules to reduce need of access to DBI or ::Connection  handles.
Make connection options as compatible possible.

Portable applications will be able to switch between two modules by using one or the other.
Portable applications must not use DBI specific methods (**dsn**, **dbh**, **sth**) or **connection** attribute.

Suggestions for the new module name?

jhthorsen:
I really wish you didn't make a new module. Unfortunately I don't have the time this week to merge your new branch with something that I think will be more compatible.

Please don't rush a new module. I really want your code.

harry-bix:
Making a module that has selectable backend is possible.
I just don't think making a DBI compatibility layer is good idea.
**dbh** and **sth** are for accessing DBI functionality if DBI backend is used, **connection** for the other.
Normally applications would not need neither.

I am still not sure if switchable backend or two separate modules is better and will continue to work on both scenarios without rushing to release new module.

jhthorsen:
Don't misunderstand me: I don't want a _complete_ compat layer with DBI. I just want to keep the "dbh" and "sth" attributes, and make them return something useful.

This might sound very stupid and I'm not sure if it is or not before trying it out myself.

My concern is that if your new shiny Connection module does not work, then the existing users have no way to get back to what used to work. That's why I would rather have a clunky API where "dbh" returns a Connection instead of just ripping it all out.

This is how it feels right to me:
- Ship one version with Connection as optional backend, and DBI as default + a warning that they will be switched around
- Ship next version with Connection as default and DBI as optional fallback.
- Ship next version without DBI. If people are happy, we can also remove all the DBI related stuff.

harry-bix:
I am thinking of similar plan.
in #11 there are some fixes and new methods that replace need of accessing 'sth' and 'dbh'.
I suggest we release #11 and next release to have EXPERIMENTAL optional Native Backend.

harry-bix:
Optional Native Backend in #12

harry-bix:
Released in separate module Mojo::MySQL5
https://github.com/harry-bix/mojo-mysql5

PR #11, opened 2015-03-06
Deprecate db, some new methods

harry-bix:
Deprecate ::Database::do in favour of query as in Mojo::Pg
Some new methods in ::Result eliminating need to access 'sth'
Bugfix in ::Migrations with trailing whitespace after last ";"


jhthorsen:
Looks great! :+1:

PR #12, opened 2015-03-10
Selectable Method for connecting to Database

harry-bix:
Mojo::mysql can use either DBI or Native implementation of MySQL client protocol.
Method is specified in Connection URL with parameter _use_dbi_. Default is true.


lib/Mojo/mysql.pm
```diff
@@ -27,62 +27,48 @@ sub db {
   delete @$self{qw(pid queue)} unless ($self->{pid} //= $$) eq $$;
 
   my ($dbh, $handle) = @{$self->_dequeue};
-  return Mojo::mysql::Database->new(dbh => $dbh, handle => $handle, mysql => $self);
+  my $db = ($self->options->{use_dbi} // 1) ?
+    Mojo::mysql::DBI::Database->new(dbh => $dbh, handle => $handle, mysql => $self) :
+    Mojo::mysql::Native::Database->new(connection => $dbh, mysql => $self);
```
jhthorsen: Why not `Mojo::mysql::Database::(Native|DBI)` ?

lib/Mojo/mysql.pm
```diff
@@ -27,62 +27,48 @@ sub db {
...
-  # especially once he discovers that DBD::mysql randomly reconnects under
-  # you, silently, but only if certain env vars are set
-  # hint: force-set mysql_auto_reconnect or whatever it's called to 0
-  $dbh->{mysql_auto_reconnect} = 0;
-  # Maintain Commits with Mojo::mysql::Transaction
-  $dbh->{AutoCommit} = 1;
-
-  $self->emit(connection => $dbh);
-  [$dbh];
+  return [undef, undef];
```
jhthorsen, line 58: I don't see how this is compatible with the return on line 67 and the code in _enqueue. Am I missing out on something?

lib/Mojo/mysql.pm
jhthorsen: Ah! I get it now, after reading more of the code :+1:

lib/Mojo/mysql.pm
harry-bix, line 58: **_enqueue** returns undef if there is no available handle to reuse.
**connect** called in **db** method if **_enqueue** returns undef.

harry-bix:
I suggest we merge this PR and make new release.

There are few things noted in #9 that are still not resolved in this PR. 
- Async connect - possible in ::Connection, not implemented in ::Database for now. Not sure which is better: connect with async callback or lazy connect.
- Transactions - Native implementation does not keep client side transaction state. DBD::mysql keeps internal state only to raise error in begin/commit/rollback if state is not correct.
  It is a good idea to have async commit/rollback. These operations can be long-running.
- Parse dsn - the only value I see here is compatibility with Mojo::Pg
- No POD in ::Native:: and ::DBI:: modules.

These issues can be addressed in next releases.

harry-bix:
@jhthorsen Do you think something else should be fixed before making a release?

jhthorsen:
I got an idea: If you rename Mojo::mysql::DBI::XXX back just Mojo::mysql::XXX we could keep the existing modules, and then when we decide to drop DBI completely, we can just rename Mojo::mysql::Native::XXX to Mojo::mysql::XXX.

What do you think @harry-bix?

jhthorsen:
I think we should mark "dsn" and "options" as deprecated in the POD, in favor of "url". I also think that many the attributes in Mojo::mysql::Connection can be replaced with "url" as well.

jhthorsen:
I would appreciate if the methods are ordered alphabetically. And "_foo" is under all the public methods.

lib/Mojo/mysql/Connection.pm
```diff
@@ -0,0 +1,797 @@
...
+  $self->{sql_state} = $self->_chew_str(5);
+  $self->{error_message} = $self->_chew_zstr;
+
+  warn '<<< ERROR ', $self->{connection_id}, ' #', $self->{seq}, ' state:', $self->_state, "\n",
+    ' error:', $self->{error_code},
+    ' state:', $self->{sql_state},
+    ' message:', $self->{error_message}, "\n" if DEBUG;
+
+  $self->_state($self->_state eq 'query' ? 'idle' : 'error');
+  $self->emit(errors => $self->{error_message});
```
jhthorsen: Why do you emit "errors" instead of "error" ?

lib/Mojo/mysql/Connection.pm
```diff
@@ -0,0 +1,797 @@
...
+    ping => '_recv_ok',
+  },
+  disconnect => {
+    idle => '_send_quit',
+    quit => '_recv_ok'
+  }
+};
+
+
+sub _flag_list($$;$) {
```
jhthorsen: Two things: It seems like a lot of these methods are "protocol" functions and not tied to $self. Could you just do `_flag_list(...)` instead of calling them as method?

The other things is that I really don't like prototypes. ($$;$) I want a really good reason to keep them around.

lib/Mojo/mysql/Connection.pm
```diff
@@ -0,0 +1,797 @@
...
+  do { _utf8_on $field->{$_} for qw(catalog schema table org_table name org_name) } if $self->options->{utf8};
+
+  push @{$self->{column_info}}, $field;
+
+  warn '<<< FIELD ', $self->{connection_id}, ' #', $self->{seq}, ' state:', $self->_state, "\n",
+    ' name:', $field->{name},
+    ' type:', REV_DATATYPE->{chr $field->{column_type}}, '(', $field->{column_type}, ')',
+    ' length:', $field->{column_length},
+    ' charset:', REV_CHARSET->{$field->{character_set}} // 'UNKNOWN', '(', $field->{character_set}, ')',
+    ' flags:', _flag_list(FIELD_FLAG, $field->{flags}), '(', $field->{flags}, ')', , "\n" if DEBUG;
```
jhthorsen: This is not very important, but I think some of these warn() statements should only be visible if DEBUG==2. (Looks like the terminal might be overloaded with information..?)

lib/Mojo/mysql/Connection.pm
```diff
@@ -0,0 +1,797 @@
...
+
+  undef $self->{$_} for qw(error_code sql_state error_message
+      affected_rows last_insert_id status_flags warnings_count field_count);
+
+  $self->{column_info} = [];
+  $self->{seq} = 0;
+  $self->{incoming} = '';
+}
+
+sub _ioloop {
```
jhthorsen: Why isn't this an attribute?

lib/Mojo/mysql/Connection.pm
```diff
@@ -0,0 +1,797 @@
...
+  }
+  else {
+    $self->$next();
+  }
+}
+
+sub _seq {
+  my ($self, $cmd, $cb) = @_;
+
+  $self->{iostream} = Mojo::IOLoop::Stream->new($self->{socket});
```
jhthorsen: Could you just call it "stream" ? More standard naming convention.

lib/Mojo/mysql/Connection.pm
```diff
@@ -0,0 +1,797 @@
...
+
+  $self->{iostream}->on(read => sub {
+    my ($stream, $bytes) = @_;
+    $self->{incoming} .= $bytes;
+
+    $self->_seq_next($cmd, 0) while $self->_seq_next_ready;
+
+    if ($self->_state eq 'idle' or $self->_state eq 'error') {
+      $stream->steal_handle;
+      delete $self->{iostream};
```
jhthorsen: Two things: Is this the same as $stream->close? And do you want to close the connection on "idle"? (Maybe I'm missing out on something here...)

lib/Mojo/mysql/Connection.pm
```diff
@@ -0,0 +1,797 @@
...
+      $cb ? $self->$cb() : $self->_ioloop(0)->stop;
+    }
+    else {
+      $self->_seq_next($cmd, 1);
+    }
+  });
+  $self->{iostream}->on(error => sub {
+    my ($stream, $err) = @_;
+    warn "stream error: $err\n" if DEBUG;
+    $self->{error_message} //= $err;
```
jhthorsen, line 467: Why don't you emit error here?

lib/Mojo/mysql/Connection.pm
```diff
@@ -0,0 +1,797 @@
...
+
+sub query {
+  my ($self, $sql, $cb) = @_;
+  $self->{sql} = $sql;
+  $self->_cmd('query', $cb);
+}
+
+sub ping {
+  my $self = shift;
+  return $self->_state eq 'disconnected' ? 0 : $self->_cmd('ping');
```
jhthorsen: What about async ping?

lib/Mojo/mysql/Connection.pm
```diff
@@ -0,0 +1,797 @@
...
+}
+
+sub DESTROY {
+  my $self = shift;
+  $self->unsubscribe($_) for qw(connect fields result end errors);
+  $self->disconnect if $self->_state eq 'idle' and $self->{socket};
```
jhthorsen, line 545: Why do you need to unsubscribe? And shouldn't disconnect "just happen" when the connection is lost to the mysql server?

lib/Mojo/mysql/Native/Database.pm
```diff
@@ -0,0 +1,227 @@
...
+use Mojo::Base 'Mojo::mysql::Database';
+
+use Mojo::mysql::Connection;
+use Mojo::mysql::Native::Results;
+use Mojo::mysql::Native::Transaction;
+use Mojo::mysql::Util qw(expand_sql parse_url);
+use Scalar::Util 'weaken';
+use Carp 'croak';
+
+has 'connection';
```
jhthorsen, line 11: Do you think calling this "dbh" will be confusing? It's a handle to the database :smile:

lib/Mojo/mysql/Native/Database.pm
```diff
@@ -0,0 +1,227 @@
...
+  return unless my $mysql = $self->mysql;
+  $mysql->_enqueue($c);
+}
+
+sub backlog { scalar @{shift->{waiting} || []} }
+
+sub begin {
+  my $self = shift;
+  $self->query('START TRANSACTION');
+  $self->query('SET autocommit=0');
```
jhthorsen, line 26: Could you rather do $self->connect->begin_work instead of those two query() calls? That way, the connection could also have state that would make begin() die if it's already in a transaction... Just like DBI.

lib/Mojo/mysql/Native/Database.pm
```diff
@@ -0,0 +1,227 @@
...
+  my $tx = Mojo::mysql::Native::Transaction->new(db => $self);
+  weaken $tx->{db};
+  return $tx;
+}
+
+sub connect {
```
jhthorsen, line 32: As mentioned, I would just pass "url" around.

lib/Mojo/mysql/Native/Database.pm
```diff
@@ -0,0 +1,227 @@
...
+  push @{$self->{waiting}}, { cb => $cb, sql => $sql, count => 0, started => 0,
+    results => Mojo::mysql::Native::Results->new };
+
+  # Blocking
+  unless ($cb) {
+    $self->connection->query($sql);
+    $self->_unsubscribe;
+    my $current = shift @{$self->{waiting}};
+    croak $self->connection->{error_message}
+      if $self->connection->{error_code} and $self->mysql->options->{RaiseError};
```
jhthorsen: I really wish we could just force RaiseError (not make it an option) and make connection die or return $err to the callback.

lib/Mojo/mysql/Native/Database.pm
```diff
@@ -0,0 +1,227 @@
...
+  });
+
+  $self->connection->on(end => sub {
+    my $c = shift;
+    return unless my $res = $self->{waiting}->[0]->{results};
+    $res->{$_} = $c->{$_} for qw(affected_rows last_insert_id warnings_count);
+  });
+
+  $self->connection->on(errors => sub {
+    my $c = shift;
```
jhthorsen, line 115: I want to make this compatible with the standard "error" event: `on(error => sub { my ($self, $err) = @_ }`

jhthorsen:
You've done some excellent hacking @harry-bix! I'm really exited about this.

Also, I might seem inconsistent in my comments above (compared to previous conversations): I've thought about it some more, and maybe making it DBI compatible is not the way to go. After all, since the user can select backends (for now) then maybe the two backends can behave differently. Yours most certainly adds some enhancement!

lib/Mojo/mysql/Connection.pm
harry-bix: Cannot remember why I did "errors" instead of "error". Changed to "error"

lib/Mojo/mysql/Connection.pm
harry-bix: All functions with prototypes are not methods. I am using prototypes only for kind of a hints.

lib/Mojo/mysql/Connection.pm
harry-bix: Packet tracing can really overload the terminal. Changed to be Enabled if DEBUG > 1

lib/Mojo/mysql/Connection.pm
harry-bix: changed to "stream"

lib/Mojo/mysql/Connection.pm
harry-bix: "stream" is destroyed on the end of a packet sequence (QUERY, PING) but socket handle is preserved. On the next sequence stream is created again.

lib/Mojo/mysql/Connection.pm
harry-bix, line 467: Added emit error

lib/Mojo/mysql/Connection.pm
harry-bix: Changed to allow async

lib/Mojo/mysql/Connection.pm
harry-bix, line 545: **unsubscribe** is normally not needed. Only if the application using it forgets to unsubscribe.
**disconnect** is a packet sequence to inform the server that client requests to end session.

lib/Mojo/mysql/Native/Database.pm
harry-bix, line 11: I really think using **dbh** for Connection is confusing. If somebody uses **dbh**, he expects DBI handle.

lib/Mojo/mysql/Native/Database.pm
harry-bix, line 26: DBI's **being_work**, **commit** and **rollback** are blocking. I think we should make exactly the opposite: Common Transaction module that uses $db->query and have async support. (in future release).
COMMIT and ROLLBACK can be long-running.

lib/Mojo/mysql/Native/Database.pm
harry-bix, line 32: I have a different plan for "url". To support failover hosts just like the JDBC connector for mysql.
http://dev.mysql.com/doc/connector-j/en/connector-j-reference-configuration-properties.html
If this functionality is added then "url" cannot be passed to Connection or Database. (_again this is future plan_).

lib/Mojo/mysql/Native/Database.pm
harry-bix: I was thinking to do exactly this. Reconsidered just to keep compatibility.
If we force RaiseError = 1 we should provide common way for Native and DBI to read extended error information (error code, error message, sql state).
If we decide to keep RaiseError option, we should provide common way for temporary disabling it. 
For DBI is:

``` perl
{
  local $db->dbh->{RaiseError} = 0;
  $db->query("...");
}
```

For Native is:

``` perl
my $saved = $db->options->{RaiseError};
$db->options->{RaiseError} = 0;
$db->query('...');
$db->options->{RaiseError} = $saved;
```

Probably an issue for next releases ...

harry-bix:
On **_send** and **_recv** methods in ::Connection
It's good to have POD for them, but I was thinking to replace them with packet sequence data definitions. That's why they remain private for now.

harry-bix:
On Renaming ::DBI:: back.
If Native modules use DBI modules as base then DBI.pm must be required for the distribution.
Using Native as base of DBI can be done.

For next releases it's good to remove dependency from DBD::mysql and change default behaviour to:
if "use_dbi" is not set then use DBI if DBD::mysql exists and have async support, else use Native.
async support for DBD::mysql is never enabled on Win32.

lib/Mojo/mysql/Connection.pm
jhthorsen: Please don't. Move it to Mojo::mysql::Util() of to the bottom of the file marked as functions as a comment.

lib/Mojo/mysql/Native/Database.pm
jhthorsen, line 11: Good enough for me :+1:

lib/Mojo/mysql/Native/Database.pm
jhthorsen, line 26: What's wrong with $self->connection->begin_work($cb); for non-blocking. It looks blocking what you have here now. Am I wrong..?

lib/Mojo/mysql/Native/Database.pm
jhthorsen, line 32: I get your point, but I still think you could pass on a _modified_ "url" object. I really find it confusing to have all of these attributes.

lib/Mojo/mysql/Native/Database.pm
jhthorsen: :+1: on just removing it _now_. Make it die() if there's an error in blocking mode. I really hate silent errors.

jhthorsen:
Renaming: I disagree. Just lazy load DBI.pm in Mojo::mysql::Database. The only reason for keeping the module naming as you have now is if we want to support DBI.pm in all of future releases. I don't think that's what we want. I think we want to remove DBI completely. (Even though that might not happen until late 2015 or in 2016)

If you think it's a good idea to keep DBI.pm around in future releases as well, then I'm happy with that and I will not try to convince you to rename it.

Another very nice side effect would be that this PR would be much easier to read. (The diff would be smaller -- that's also why I opened #13)

lib/Mojo/mysql/Connection.pm
jhthorsen: :+1:

lib/Mojo/mysql/Connection.pm
jhthorsen: Thanks.

lib/Mojo/mysql/Connection.pm
jhthorsen: Could you point me to where the handle is preserved? https://metacpan.org/pod/Mojo::IOLoop::Stream#steal_handle returns $handle, but I don't see where you store it...

lib/Mojo/mysql/Connection.pm
jhthorsen, line 467: :+1:

lib/Mojo/mysql/Connection.pm
jhthorsen: Awesome!

lib/Mojo/mysql/Connection.pm
jhthorsen, line 545: Why do you need to unsubscribe if the application forgets to unsubscribe? Perl should just garbage collect all of this..?

lib/Mojo/mysql/Native/Database.pm
harry-bix, line 26: Yes, it is blocking for now.
This is a minimal implementation compatible with ::DBI::Transaction.
My idea is to make common ::Transaction module that uses query and async support in future release. This will need API changes.

jhthorsen, line 26: Will these commands fail if the server is already in a transaction? What I want is that calling "begin()" should fail the second time, unless you have issued rollback or commit.

harry-bix, line 26: DBI keeps internal state and fails if second begin_work is called.
Native ::Transaction does not. (for now)
Second **START TRANSACTION** does implicit **COMMIT** and MySQL does not complain.

lib/Mojo/mysql/Connection.pm
harry-bix: handle stays in $connection->{socket}
steal_handle is needed to prevent Stream DESTROY to close the handle

lib/Mojo/mysql/Native/Database.pm
jhthorsen, line 26: I think implicit COMMIT is a bad idea that need to be fixed before release.

harry-bix, line 26: Fixed to check status_flags if already in transaction.
For **COMMIT** and **ROLLBACK** this check is not necessary.

lib/Mojo/mysql/Native/Database.pm
harry-bix, line 32: Using **url** everywhere where we need connect info is not a bad idea.
If we do this we have to either deprecate **username** and **password** in mysql.pm
or update the url when username or password is set.

lib/Mojo/mysql/Native/Database.pm
harry-bix: I also do not like RaiseError=0 and never use it, but enforcing it will be incompatible change. Lets keep it for now.

lib/Mojo/mysql/Native/Database.pm
jhthorsen, line 32: Deprecate "username" and "password" :+1:

lib/Mojo/mysql/Native/Database.pm
jhthorsen: I don't get it... we're two against zero on this vote. What is stopping us? :smiley_cat:

lib/Mojo/mysql/Native/Database.pm
harry-bix, line 32: Deprecated dsn, username and password. Deprecation warning only when used as setter.
Not sure if url should be updated.

jhthorsen:
Have a look here about RaiseError and AutoCommit: https://github.com/kraih/mojo-pg/commit/fe4c3ae84015bf0ed87cc2fd2de6b43d9a2607bd. I guess that means that you should remove it completely from the native code.

lib/Mojo/mysql/Native/Database.pm
harry-bix: Done. With pleasure.

harry-bix:
Done for RaiseError. For AutoCommit we already did it. Nice to have the same behaviour with Mojo::Pg.

harry-bix:
I've deprecated options also. Not sure where options defaults should be.

harry-bix:
Released in separate module Mojo::MySQL5
https://github.com/harry-bix/mojo-mysql5

PR #14, opened 2015-03-26
migrations parser

harry-bix:
- support for delimiter
- correct split of sql statements
- ';' or other defined delimiter in quotes and comments
- possible to define stored procedures and triggers in migration scripts

### DIFF
--- a/lib/Mojo/mysql/Migrations.pm
+++ b/lib/Mojo/mysql/Migrations.pm
@@ -23,11 +23,76 @@ sub from_file { shift->from_string(decode 'UTF-8', slurp pop) }
 sub from_string {
   my ($self, $sql) = @_;
   my ($version, $way);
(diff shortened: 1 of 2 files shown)


jhthorsen:
What is the reason for this PR? What does it add?

harry-bix:
It is a bugfix and enhancement.
Migration scripts were failing if the delimiter (;) was contained in quotes or comments.
The enhancement is the custom **delimiter** command just like in the mysql client.
**delimiter** command allows definition of stored procedures or triggers in migration scripts.
More complex parser was needed in place of split(';', $sql)

harry-bix:
https://dev.mysql.com/doc/refman/5.6/en/mysql-commands.html

https://dev.mysql.com/doc/refman/5.6/en/stored-programs-defining.html

jhthorsen:
Cool. Thanks for explaining.

PR #16, opened 2015-04-03
Remove discouraging text about migrations

harry-bix:
Migrations are really nice feature.
The current POD looks like they are something very dangerous.
It's better to explain possible unexpected behaviour of the MySQL engine.

### DIFF
--- a/lib/Mojo/mysql.pm
+++ b/lib/Mojo/mysql.pm
@@ -249,7 +249,7 @@ easily.
 
 MySQL does not support nested transactions and DDL transactions.
 DDL statements cause implicit C<COMMIT>. C<ROLLBACK> will be called if
-any step of migration script fails, but not everything can be reverted.
+any step of migration script fails, but only DML statements can be reverted.
 
 This means database can be left in unknown state if migration script fails.
 Use this feature with caution and remember to always backup your database.
--- a/lib/Mojo/mysql.pm
+++ b/lib/Mojo/mysql.pm
@@ -253,7 +253,7 @@ any step of migration script fails, but only DML statements after the
 last implicit or explicit C<COMMIT> can be reverted.
 Not all MySQL storage engines (like C<MYISAM>) support transactions.
 
-This means database can be left in unknown state if migration script fails.
+This means database will most likely be left in unknown state if migration script fails.
 Use this feature with caution and remember to always backup your database.
 
 =head2 options
--- a/lib/Mojo/mysql.pm
+++ b/lib/Mojo/mysql.pm
@@ -249,7 +249,10 @@ easily.
 
 MySQL does not support nested transactions and DDL transactions.
 DDL statements cause implicit C<COMMIT>. C<ROLLBACK> will be called if
-migration script fails, but not everything will be reverted.
+any step of migration script fails, but not everything can be reverted.
+
+This means database can be left in unknown state if migration script fails.
+Use this feature with caution and remember to always backup your database.
 
 =head2 options
 
--- a/lib/Mojo/mysql.pm
+++ b/lib/Mojo/mysql.pm
@@ -238,10 +238,6 @@ C<5>.
 
 =head2 migrations
 
-MySQL does not support nested transactions and DDL transactions.
-DDL statements cause implicit C<COMMIT>.
-B<Therefore, migrations should be used with extreme caution. Backup your database. You've been warned.> 
-
   my $migrations = $mysql->migrations;
   $mysql         = $mysql->migrations(Mojo::mysql::Migrations->new);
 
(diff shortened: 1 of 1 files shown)


jhthorsen:
I don't agree. I really think we should inform people that their database will be (in many cases) messed up if the migration fail.

harry-bix:
Migration failure can leave the database in unwanted state. And we should note this in the POD.
But this POD is for developers, not dbadmins or sysadmins.
And written in this way sounds discouraging for using the feature.

jhthorsen:
But why shouldn't we discourage something that is so fragile?

harry-bix:
Applications do need to change the database schema. This feature gives a possible way to do this.
Fragile is the transaction model of current MySQL engine, not the feature itself.

lib/Mojo/mysql.pm
```diff
@@ -251,6 +247,10 @@ easily.
   # Load migrations from file and migrate to latest version
   $mysql->migrations->from_file('/Users/sri/migrations.sql')->migrate;
 
+MySQL does not support nested transactions and DDL transactions.
+DDL statements cause implicit C<COMMIT>. C<ROLLBACK> will be called if
+migration script fails, but not everything will be reverted.
```
jhthorsen: What will be ROLLBACK? And what is "not everything"? I can accept a change if we point the reader to another resource (or give more details) on what is really going on.

harry-bix:
The shortest explanation is here: https://dev.mysql.com/doc/refman/5.0/en/cannot-roll-back.html
The reason why is already in the POD.

harry-bix:
Detailed explanation for most popular databases:
https://wiki.postgresql.org/wiki/Transactional_DDL_in_PostgreSQL:_A_Competitive_Analysis

jhthorsen:
Then (at least to me) the documentation is now misleading: It sounds like some statements _could_ be rolled back.

We need to be explicit about this. How about:

```
MySQL does not support nested transactions and DDL transactions. This means
that your database will be in an unknown state if any migrations step fail. Use this
feature with caution and remember to always backup your MySQL database.
DDL statements will also cause implicit C<COMMIT>.
```

Maybe even with "See also https://dev.mysql.com/doc/refman/5.0/en/cannot-roll-back.html"

harry-bix:
In the last push:
Explanation of MySQL transaction specifics in first paragraph, warning in second.

harry-bix:
Mojo::mysql::Migration executes migration script steps in transaction.
If one step fails it does not execute following steps and and issues ROLLBACK. (undef $tx)
Migration scripts can contain DML statements also. At least they can be reverted.

harry-bix:
Last push clarifies that only DML statements can be rolled-back.

jhthorsen:
I think we have a bug: Why is there a begin/commit in https://github.com/jhthorsen/mojo-mysql/blob/9eb7880b1acca9ce353b0549bdd2fd48fed0e84f/lib/Mojo/mysql/Migrations.pm#L104, when the first DDL statement will issue a commit anyway? We should remove the begin/commit part if that is true. I find it very unlikely that a migration script does not contain a DDL statement.

jhthorsen:
I think we should add CAVEATS (or something) from Mojo::mysql::Transaction and then link from ::Migrations to ::Transaction:
- https://dev.mysql.com/doc/refman/5.0/en/cannot-roll-back.html
- http://dev.mysql.com/doc/refman/5.5/en/implicit-commit.html
- http://dev.mysql.com/doc/internals/en/transactions-notes-on-ddl-and-normal-transaction.html
- https://wiki.postgresql.org/wiki/Transactional_DDL_in_PostgreSQL:_A_Competitive_Analysis#MySQL

harry-bix:
It's not a bug that we execute migration statements in transaction.
A call to $dbh->begin_work, issues **START TRANSACTION** and **SET autocommit=0**.
$dbh->rollback, issues **ROLLBACK** and restores **autocommit** to it's state before begin_work.

So on migration script failure, data modified by DML statements after the last implicit or explicit **COMMIT** will be reverted because **autocommit** is turned off.

jhthorsen:
But it doesn't matter, does it..?

```
$dbh->begin_work;
$dbh->do("DELETE * FROM some_table");
$dbh->do("ALTER TABLE other_table ADD COLUMN ..."); # implicit commit
$dbh->rollback; # doesn't matter
```

harry-bix:
Yes, but there is no point in removing migrations in transaction.
We have correct explanation of what can happen.
Knowing that developer can design his migration scripts to be as predictable as possible.

jhthorsen:
@harry-bix: Can you join #mojo on irc.perl.org ?

harry-bix:
there as 'harry'

jhthorsen:
Ok, so after discussing on IRC, I agree. Especially after the details around regarding how the database is still in a transaction if...

> MySQL keeps two different states for AUTOCOMMIT and STATUS_IN_TRANS. After implicit COMMIT STATUS_IN_TRANS is 0, but since STATUS_AUTOCOMMIT is off it acts as in transaction.

PR #19, opened 2015-04-28
PubSub for MySQL

harry-bix:
MySQL does not have LISTEN/NOTIFY mechanism like PostgreSQL.
This implementation assigns single database connection to wait for notification by executing **SLEEP** on sever. Notifying connections wake up listening connections by executing **KILL QUERY**. Notify payload and list of subscribed connections is stored in tables.

Works on MySQL 5.0 and above.

### DIFF
--- a/lib/Mojo/mysql/PubSub.pm
+++ b/lib/Mojo/mysql/PubSub.pm
@@ -3,38 +3,56 @@ use Mojo::Base 'Mojo::EventEmitter';
 
 use Scalar::Util 'weaken';
 
+use constant DEBUG => $ENV{MOJO_PUBSUB_DEBUG} || 0;
+
 has 'mysql';
 
 sub listen {
-  my ($self, $name, $cb) = @_;
-  my $sleeping = $self->_db;
-  my $sql = @{$self->{chans}{$name} ||= []} ?
+  my ($self, $channel, $cb) = @_;
+  my $pid = $self->_db->pid;
+  warn "listen $channel listening:$pid\n" if DEBUG;
+  $self->mysql->db->query(@{$self->{chans}{$channel} ||= []} ?
     'update mojo_pubsub_subscribe set ts = current_timestamp where pid = ? and channel = ?' :
-    'insert into mojo_pubsub_subscribe(pid, channel) values (?, ?)';
-  $self->mysql->db->query($sql, $sleeping->pid, $name);
-  push @{$self->{chans}{$name}}, $cb;
+    'insert into mojo_pubsub_subscribe(pid, channel) values (?, ?)', $pid, $channel);
+  push @{$self->{chans}{$channel}}, $cb;
   return $cb;
 }
 
 sub notify {
-  my ($self, $name, $payload) = @_;
-  $self->_db;
+  my ($self, $channel, $payload) = @_;
+  $payload //= '';
+  my $pid = $self->_db->pid;
+  warn "notify $channel $payload listening:$pid\n" if DEBUG;
   $self->mysql->db->query(
-    'insert into mojo_pubsub_notify(channel, message) values (?, ?)', $name, $payload);
+    'insert into mojo_pubsub_notify(channel, payload) values (?, ?)', $channel, $payload);
   return $self;
 }
 
 sub unlisten {
-  my ($self, $name, $cb) = @_;
-  my $chan = $self->{chans}{$name};
+  my ($self, $channel, $cb) = @_;
+  my $pid = $self->_db->pid;
+  warn "unlisten $channel listening:$pid\n" if DEBUG;
+  my $chan = $self->{chans}{$channel};
   @$chan = grep { $cb ne $_ } @$chan;
   return $self if @$chan;
-  my $sleeping = $self->_db;
   $self->mysql->db->query(
-    'delete from mojo_pubsub_subscribe where pid = ? and channel = ?',
-    $sleeping->pid, $name);
-  delete $self->{chans}{$name};
+    'delete from mojo_pubsub_subscribe where pid = ? and channel = ?', $pid, $channel);
+  delete $self->{chans}{$channel};
   return $self;
+}
+
+sub _recv_notifications {
+  my $self = shift;
+  my $result = $self->{db}->query(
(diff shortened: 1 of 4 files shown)


jhthorsen:
This is so evil (read: creative) that I just want to hit the _Merge_ button right away :smile_cat: 

I need to read the code twice though before I merge it though. I'll be back as soon as possible with comments.

jhthorsen:
I've been going back and forth on the table names... Could we come up with something without the "mojo_" prefix? Like just "pubsub_" or something longer like "publish_and_subscribe_" or "pubsub_compat_layer_". I was also thinking this could be configurable, but I don't care about that anymore.

lib/Mojo/mysql/PubSub.pm
```diff
@@ -0,0 +1,291 @@
...
+      delete $self->{db};
+      eval { $self->_db };
+    }
+    elsif ($self and $self->{db}) {
+      $self->_notifications;
+      $db->query('update mojo_pubsub_subscribe set ts = current_timestamp where pid = ?', $db->pid);
+      $db->query('select sleep(600)', $cb);
+    }
+  };
+  $db->query('select sleep(600)', $cb);
```
jhthorsen: Does this mean that both the subscriber and publisher are running the `sleep(600)` query? I think I would do `$self->_db($bool)` to indicate if this is a subscriber or not or simply refactor this method. Let me know if I'm missing out on something.

lib/Mojo/mysql/PubSub.pm
```diff
@@ -0,0 +1,291 @@
...
+
+=cut
+
+__DATA__
+
+@@ pubsub
+-- 1 up
+drop table if exists mojo_pubsub_subscribe;
+drop table if exists mojo_pubsub_notify;
+drop trigger if exists mojo_pubsub_notify_kill;
```
jhthorsen: Why do you drop anything? I'm thinking these tables shouldn't exists, and if they do, I would be very cautious about dropping them.

lib/Mojo/mysql/PubSub.pm
```diff
@@ -0,0 +1,291 @@
...
+  channel varchar(64) not null,
+  ts timestamp not null default current_timestamp,
+  unique key subs_idx(pid, channel),
+  key ts_idx(ts)
+);
+
+create table mojo_pubsub_notify(
+  id integer auto_increment primary key,
+  channel varchar(64) not null,
+  payload varchar(256),
```
jhthorsen: Why can't payload be TEXT?

lib/Mojo/mysql/PubSub.pm
```diff
@@ -0,0 +1,291 @@
...
+      kill query t_pid;
+    end if;
+
+  until done end repeat;
+
+  close subs_c;
+end
+//
+
+delimiter ;
```
jhthorsen: Shouldn't there be a ```-- 1 down``` step as well?

lib/Mojo/mysql/PubSub.pm
harry-bix: Only the subscriber is sleeping, sleeps again after wake up
The outside sleep can be replaced with $db->$cb();

harry-bix:
Thinking about table names
Mojo::mysql::Migrations uses **mojo_migrations** table.
Mojo::mysql::PubSub uses **mojo_pubsub** prefix for its tables.

jhthorsen:
Yeah. Sounds very consistent. Let's keep the table names as is.

lib/Mojo/mysql/PubSub.pm
harry-bix: This will be executed only if there is no 'pubsub' record in mojo_migrations table.
t/migrations.t re-creates mojo_migrations table, and pubsub.t tries to execute pubsub migration again. Without dropping pubsub migration will fail, so all unit tests cannot be run more than once on same database.

jhthorsen: That's why I asked you to add the down step as well. That way, you can downgrade the schema first in t/pubsub.t:

```
$mysql->migrations->name('pubsub')->from_data('Mojo::mysql::PubSub')->migrate(0);
```

t/pubsub.t
```diff
@@ -0,0 +1,79 @@
...
+  Mojo::IOLoop->start;
+  ok $dbhs[1], 'database handle';
+  isnt $dbhs[0], $dbhs[1], 'different database handles';
+  Mojo::IOLoop->delay(
+    sub { Mojo::IOLoop->timer(0.05, shift->begin); },
+    sub { is_deeply \@test, ['works'], 'right messages'; }
+  )->wait;
+};
+
+$mysql->migrations->name('pubsub')->from_data('Mojo::mysql::PubSub')->migrate(0);
```
jhthorsen, line 63: :+1:

lib/Mojo/mysql/PubSub.pm
```diff
@@ -0,0 +1,289 @@
...
+  channel varchar(64) not null,
+  payload text,
+  ts timestamp not null default current_timestamp,
+  key channel_idx(channel),
+  key ts_idx(ts)
+);
+
+delimiter //
+
+create trigger mojo_pubsub_notify_kill after insert on mojo_pubsub_notify
```
jhthorsen, line 277: I'm curious if there's any special GRANT that is required for this pubsub code to work:
- Can anyone select from INFORMATION_SCHEMA.PROCESSLIST?
- Can a user kill a query that another user is running?
- Should this procedure be inserted by a user with higher database rights, than users using this module?

harry-bix, line 277: - **PROCESS** privilege to see INFORMATION_SCHEMA.PROCESSLIST for other users.
- **SUPER** privilege for **KILL QUERY** started by other user
- **SUPER** may be required to define trigger (if binary logging is enabled)
- only **SELECT,INSERT,UPDATE,DELETE** on mojo_pubsub_subscribe and mojo_pubsub_notify tables for normal use of the module.

Generally it good to execute the 'pubsub' migration by user having **SUPER** privilege on database.

jhthorsen:
I'm happy @harry-bix! Merge when you feel like it :+1:

PR #22, opened 2016-01-17
fixed "(in cleanup) Can't call method "FETCH" on an undefined value"

kberov:
...during global destruction.


jhthorsen:
Thanks. I will have a look at this next week, unless @harry-bix beats me to it.

kberov:
:+1: Thanks, I am not sure tough while the tests are failing, Looks like some configuration stuff rather than real fails

jhthorsen:
Pretty sure it's real fails. Seing the same thing locally with the patch.

jhthorsen:
Can you provide an example (test?) where this error is seen?

jhthorsen:
Closing this because of inactivity.

PR #24, opened 2016-01-27
Use prepare_cached from DBI to cache statement handles

jhthorsen:
I would like to pull in https://github.com/kraih/mojo-pg/commit/42e329ef37fe117d43c1c0929a027fb1498e8d03 which use prepare_cached() instead of just prepare() to speed up queries.

But database.t seems to hang with the change. Any idea why @harry-bix ?

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,5 +1,9 @@
 Revision history for perl distribution Mojo-mysql
 
+0.14 Not Released
+ - Use prepare_cached from DBI to cache statement handles
+   https://github.com/kraih/mojo-pg/commit/42e329ef37fe117d43c1c0929a027fb1498e8d03
+
 0.13 2016-01-27T21:05:37+0100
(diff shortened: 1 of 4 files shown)


harry-bix:
MySQL statement handles cannot be re-used.
**prepare_cached** is implemented in DBI, not DBD::mysql.
It can be useful only if server-side prepared statements are used, but Mojo::mysql fails if mysql_server_prepare=1 is used. (not because of a bug in it).

jhthorsen:
Does that mean we can close this as "wontfix"?

What about this commit? https://github.com/kraih/mojo-pg/commit/09e0ba4f6331f6ac3d3b719280fb4ad6b44cac0b Or is simply "prepare" useless in MySQL context?

jhthorsen:
Thanks :smiley:

PR #26, opened 2016-08-02
Update mojo_migrations key type

Akron:
When creating the `mojo_migrations` table, the creation may fail with the warning `DBD::mysql::st execute failed: Specified key was too long; max key length is 767 bytes`. The reason is, that the key is specified as a `varchar(255)` that, in a utf8 world (that is the default in some databases) may exceed the 767 limit. There seems to be a lot of discussion about this limitation and there is now an option to increase this limit (`innodb_large_prefix`), but I guess, limiting the size of the key field is a reasonable fix. Another possible fix would be to change the encoding of the table, but I've chosen the varbinary-option as the limitation is byte-bound and not character set bound.

I am no database expert - so I may be totally wrong. ;)
Thanks for your great work on this module!
Best,
Akron

My database is:
Ver 15.1 Distrib 10.0.25-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2

### DIFF
--- a/lib/Mojo/mysql/Migrations.pm
+++ b/lib/Mojo/mysql/Migrations.pm
@@ -146,7 +146,7 @@ sub _active {
 
   $db->query(
     'create table if not exists mojo_migrations (
-       name    varchar(255) unique not null,
+       name    varbinary(255) unique not null,
        version bigint not null
      )'
   ) if $error;


harry-bix:
**varchar(255)** fits in **innodb_large_prefix** of 767 then the encoding is **utf8**.
It does not fit if encoding is **utf8mb4**.

Changing name datatype to varbinary will prevent possibility to name migration in non-ascii languages.

Akron:
Ah - I wasn't aware of the limitation of the `utf8` called charset. You are right. But it seems, that there are some installations of MySQL/MariaDB where a max-4byte-length encoding is the default. My MariaDB installation at least was fresh and complained about the key length. I don't think, that the varbinary change prevents any non-ascii migration names (it is just up to the developer to correctly encode/decode the name) - but you would prefer to set the charset to the MySQL `utf8`?

harry-bix:
It's best to leave the character set to developer or db admin.
I suggest to replace **varchar(255)** with **varchar(128)**.
It will be long enough and work with 4-byte character sets.

jhthorsen:
What happens with this PR? I'm no expert either, so I need someone with the qualifications to say "go".

Grinnz:
I think @harry-bix's suggestion is reasonable. Set the column size smaller so that it supports 4 bytes as that is the largest mysql supports.

jhthorsen:
Ok! Closing this with 41a3685 then.

Thanks everybody.

PR #27, opened 2017-02-17
fixed method references in perldoc

fheyer:
Fixed references to method abstract in Mojo::mysql - obviously left over from porting from Mojo::Pg ;)
Great idea to add SQL::Abstract support!


jhthorsen:
Wow! Thanks 👍 

Copy/paste ftw, hehe

PR #28, opened 2017-03-14
implemented Mojo::mysql::Results::map and ::map_hashes

rndev15:


jhthorsen:
What benefit does this add?

rndev15:
Easy way to map result by key.
`$mysql->db->query('select * from posts')->map_hashes('id);`
{ 1 => { name => 'foo', id => 1}, ...}
`$mysql->db->query('select * from posts')->map;`
{ 1 => 'foo', ... }

lib/Mojo/mysql/Results.pm
```diff
@@ -18,6 +18,24 @@ sub hash { shift->sth->fetchrow_hashref }
 
 sub hashes { Mojo::Collection->new(@{shift->sth->fetchall_arrayref({})}) }
 
+sub map {
+  return { map @$_, @{ $_[0]->arrays } };
```
jhthorsen, line 22: How does this work if you select three columns?

lib/Mojo/mysql/Results.pm
```diff
@@ -18,6 +18,24 @@ sub hash { shift->sth->fetchrow_hashref }
...
+  my @keys;
+
+  $rows->map(sub {
+    push @keys, $_->{ $keyname };
+  });
+
+  my %return;
+  @return{@keys} = @$rows;
+  return \%return;
+}
```
jhthorsen, line 37: This is a lot of code. I would much rather run the code below, without the need of an extra method in Mojo::mysql::Results:

    %mapped = $res->hashes->map(sub { ($_->{id} => $_) })->each;

jhthorsen (commented):
This PR has no documentation, so it's hard to understand what you're trying to accomplish. But I won't take it, if I have understood it correctly, since there are already quite simple ways to accomplish this.

rndev15:
Ok, I have understand you. May be you are right, it's not necessary.
Sorry for disturb.

jhthorsen:
No worries! Thanks for the interest 👍

PR #31, opened 2017-05-19
Add ability to set types of query parameters

Grinnz:
This is required as binary data must be specified as SQL_BLOB in order to not be UTF-8 encoded, while using the fixed mysql_enable_utf8/mb4 feature (in DBD::mysql 4.042 which Mojo::mysql already depends on). I don't have a running mysql server to run the tests locally so please verify that they work.


kraih:
`Mojo::Pg` does this as well. 👍

jhthorsen:
Awesome! Thanks for the pull request 👍

PR #32, opened 2017-05-19
update mysql_enable_utf8 option to mysql_enable_utf8mb4

Grinnz:
The mysql_enable_utf8 sets the utf8 charset which only supports up to 3-byte UTF-8 encodings. mysql_enable_utf8mb4 (as of DBD::mysql 4.032) properly supports encoding unicode characters to up to 4 bytes, such as 𠜎. It means the connection charset will be utf8mb4 (supported back to at least mysql 5.5) and these unicode characters will be supported, but no other changes.

### DIFF
--- a/lib/Mojo/mysql.pm
+++ b/lib/Mojo/mysql.pm
@@ -20,7 +20,7 @@ has migrations      => sub {
   return $migrations;
 };
 has options => sub {
-  {mysql_enable_utf8 => 1, AutoCommit => 1, AutoInactiveDestroy => 1, PrintError => 0, RaiseError => 1};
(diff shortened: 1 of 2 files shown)


Grinnz:
If you want to still support mysql 5.1, then this will have to be optional somehow.

jhthorsen:
I'm not sure about the consequences here.

Could you provide a failing test where data is not inserted correctly without this setting?

akarelas:
If inserted string contains a 4-byte emoji, the string gets truncated at that point and saved, with no error I believe. I'm letting Grinnz provide the test.

jhthorsen:
I would say it's quite easy to change that when creating a new object, so maybe it should just be documented at this point to support older MySQL versions?

Grinnz:
This is true. Documentation and a recommendation in the synopsis would be a simple way.

jhthorsen:
Closing this after 558cb7c.

PR #36, opened 2018-02-22
Accept Mojo::URL object as input for database URL

krunen:
If you send a Mojo::URL object with userinfo to Mojo::mysql->new(), the userinfo will be stripped by Mojo::URL->new(). The stringification of Mojo::URL will strip it. This change will accept Mojo::URL obects as input to new() and use it as is.

### DIFF
--- a/lib/Mojo/mysql.pm
+++ b/lib/Mojo/mysql.pm
@@ -49,7 +49,7 @@ sub from_string {
 
   # Protocol
   return $self unless $str;
-  my $url = Mojo::URL->new($str);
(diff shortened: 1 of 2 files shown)


jhthorsen:
Can you also write a test for this?

krunen:
I added a test for the _new_ method with a URL obj as param.
Is the _from_string_ method a bit misnamed now? Pedantically, maybe the test for a URL object should be in the _new_ method instead?

jhthorsen:
I think it's fine, since this is probably more DWIM. I would be opposed if the URL object did not stringify.

Nice work! 👍

PR #37, opened 2018-05-02
allow constructor to take a single hashref

preaction:
The documentation says this works, but it didn't.

### DIFF
--- a/lib/Mojo/mysql.pm
+++ b/lib/Mojo/mysql.pm
@@ -73,7 +73,9 @@ sub from_string {
   return $self->dsn($dsn);
 }
 
-sub new { @_ == 2 ? shift->SUPER::new->from_string(@_) : shift->SUPER::new(@_) }
+sub new {
+  @_ > 2 || ref $_[-1] eq 'HASH' ? shift->SUPER::new(@_) : shift->SUPER::new->from_string(@_);
+}
 
 sub strict_mode {
   my $self = ref $_[0] ? shift : shift->new(@_);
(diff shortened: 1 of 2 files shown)


lib/Mojo/mysql.pm
```diff
@@ -7,7 +7,7 @@ use File::Spec::Functions 'file_name_is_absolute';
 use Mojo::mysql::Database;
 use Mojo::mysql::Migrations;
 use Mojo::URL;
-use Scalar::Util 'weaken';
+use Scalar::Util 'weaken', 'blessed';
```
jhthorsen: What does `blessed` do? Looks like it's extra..?

lib/Mojo/mysql.pm
```diff
@@ -73,7 +73,11 @@ sub from_string {
   return $self->dsn($dsn);
 }
 
-sub new { @_ == 2 ? shift->SUPER::new->from_string(@_) : shift->SUPER::new(@_) }
+sub new {
+    @_ > 2 || ref $_[-1] eq 'HASH'
+        ? shift->SUPER::new(@_)
+        : shift->SUPER::new->from_string(@_)
```
jhthorsen: Please tidy both files, if you want `git blame` to reference you later on 😄

jhthorsen (changes requested):
Wow! What a dumb mistake of me. Thanks for the patch 👍

lib/Mojo/mysql.pm
preaction: Sorry, yes, as I was iterating to find the solution I could accept, I had imported `blessed` to try to distinguish between Mojo::URL objects and plain hashrefs (which it turns out `ref` handles just fine). I'll remove this.

lib/Mojo/mysql.pm
preaction: Done!

jhthorsen:
Cool! I’ll try to make a release when I get to work in a couple of hours.

PR #44, opened 2018-12-01
Skip pubsub tests if TEST_PUBSUB is not set

Tekki:


jhthorsen:
Can you update .travis.yml as well, so the test will be automatically tested?

Tekki:
BTW shouldn't we test Perl 5.28 on Travis, too?

jhthorsen:
You can open up a PR if you think we should test on 5.28 👍

PR #46, opened 2018-12-01
Add SQL::Abstract::mysql

Tekki:


lib/SQL/Abstract/mysql.pm
```diff
@@ -0,0 +1,263 @@
...
+  else {
+    $command = 'insert';
+  }
+  $sql = join ' ', $self->_sqlcase("$command into"), $table, $sql;
+
+  return wantarray ? ($sql, @bind) : $sql;
+}
+
+sub new {
+  return shift->SUPER::new(quote_char => chr(96), name_sep => '.');
```
jhthorsen: Why did you move this here instead of keeping it in Mojo::mysql? Shouldn't it be possible to customise this?

t/json.t
```diff
@@ -26,6 +27,4 @@ is_deeply $db->query('select name from mojo_json_test where name like "%supergir
 is_deeply $db->query('select name from mojo_json_test where name like "%supergirl%"')->expand(1)->hash,
   {name => {nick => 'supergirl'}}, 'name as hash';
 
-$db->query('drop table mojo_json_test');
```
jhthorsen, line 29: I don't understand this change. Why don't you want to cleanup after the test is run? Also, why not use `if not exists` above?

jhthorsen (commented):
Looks fine in general 👍

lib/SQL/Abstract/mysql.pm
Tekki: We could leave it the same way as Mojo::Pg does, but then we should document it.

t/json.t
Tekki: Why I don't want cleanup: Me personally I find it convenient to have tables containing some data after the tests.
Why no `if not exists`: If the tests fail no cleanup is made and the next time the tests start with a table already containing some data. Because of that the test `select id, name, j from mojo_json_test where json_extract(j, "$.foo") = 42` will not return `$$` as ID, but something else.

lib/SQL/Abstract/mysql.pm
jhthorsen: I would appreciate that. Or at least pass in `@_` so the end user can override the settings.

t/json.t
jhthorsen: I will allow this:

```
$db->query('drop table mojo_json_test') unless $ENV{TEST_KEEP_DB};
```

lib/SQL/Abstract/mysql.pm
Tekki: I'll move it back and update the docs.

t/json.t
Tekki: Changed at 2 places and added 'truncate table' commands at the beginning.

jhthorsen:
There are some issues with tidying, but I'll take it anyways and fix it locally.

Thanks for the PR! 👍

Tekki:
Seems I forgot to run perltidy at some places.
Thanks for merging!

jhthorsen:
No problem 👍Thanks for your contribution! I updated the documentation a couple of places as well. Hope I didn't mess up anything. https://github.com/jhthorsen/mojo-mysql/commit/fa98faeced16bb4c78fc6d3d32f08973ca80d2a5

https://metacpan.org/pod/SQL::Abstract::mysql

lib/SQL/Abstract/mysql.pm
```diff
@@ -0,0 +1,263 @@
...
+
+  # "insert into t (id, a) values (123, 'b') on duplicate key update c='d'"
+  $abstract->insert('t', {id => 123, a => 'b'}, {on_conflict => {c => 'd'}});
+
+=head2 select
+
+  $abstract->select($source, $fields, $where, $order);
+  $abstract->select($source, $fields, $where, \%options);
+
+=head3 AS
```
jhthorsen, line 160: @Tekki: I can't see where this is implemented. Was it planned, but then not finished..? I'm a bit surprised I didn't see this earlier.

Tekki: Yes, `sub _select_fields` is missing. I think the following happened: I started with the code from SQL::Abstract::Pg, then saw thought this method is not compatible with the JSON support I wanted to add, so I 'temporarily' removed it.
The question is what we do now, remove this paragraph from the docs or add the method?

jhthorsen: I would like to have this functionality :)

Tekki: Here is the PR: https://github.com/jhthorsen/mojo-mysql/pull/69

PR #47, opened 2018-12-05
Add support for mariadb:// and DBD::MariaDB

jhthorsen:
I'm not entirely sure if there should be a Mojo::MariaDB or if I could add support for MariaDB directly into Mojo::mysql.

Keeping this PR open for input from the public for now.

### DIFF
--- a/lib/Mojo/mysql.pm
+++ b/lib/Mojo/mysql.pm
@@ -25,7 +25,7 @@ has migrations => sub {
 };
 
 has options => sub {
-  my $self = shift;
+  my $self    = shift;
   my $options = {AutoCommit => 1, AutoInactiveDestroy => 1, PrintError => 0, RaiseError => 1};
   $options->{mysql_enable_utf8} = 1 if $self->dsn =~ m!^dbi:mysql:!;
   return $options;
@@ -68,6 +68,10 @@ sub from_string {
   my $url = UNIVERSAL::isa($str, 'Mojo::URL') ? $str : Mojo::URL->new($str);
   croak qq{Invalid MySQL/MariaDB connection string "$str"} unless $url->protocol =~ m!^(mariadb|mysql)$!;
   my $dsn = $url->protocol eq 'mariadb' ? 'dbi:MariaDB' : 'dbi:mysql';
+
+  # https://github.com/jhthorsen/mojo-mysql/pull/47
+  die "DBD::MariaDB 1.21 is required for Mojo::mysql to work properly"
+    if $dsn eq 'dbi:MariaDB' and !eval 'use DBD::MariaDB 1.21;1';
 
   # Database
   $dsn .= ':dbname=' . $url->path->parts->[0] if defined $url->path->parts->[0];
@@ -473,8 +477,8 @@ This method is EXPERIMENTAL and could be removed in future releases.
 Construct a new L<Mojo::mysql> object either from L</ATTRIBUTES> and or parse
 connection string with L</"from_string"> if necessary.
 
-Using the "mariadb" scheme requires the optional module L<DBD::MariaDB> to be
-installed.
+Using the "mariadb" scheme requires the optional module L<DBD::MariaDB> version
+1.21 (or later) to be installed.
 
 =head2 strict_mode
 
--- a/lib/Mojo/mysql.pm
+++ b/lib/Mojo/mysql.pm
@@ -96,14 +96,6 @@ sub from_string {
   return $self;
 }
 
-sub handle_attr {
-  my ($self, $handle) = (shift, shift);
-  my $key = $self->dsn =~ m!^dbi:(\w+):! ? lc "$1_$_[0]" : "mysql_$_[0]";
-  return $handle->{$key} if @_ == 1;
-  $handle->{$key} = $_[1];
-  $handle;
-}
-
 sub new {
   @_ > 2 || ref $_[-1] eq 'HASH' ? shift->SUPER::new(@_) : shift->SUPER::new->from_string(@_);
 }
@@ -114,6 +106,14 @@ sub strict_mode {
   warn "[Mojo::mysql] strict_mode($self->{strict_mode})\n" if $ENV{DBI_TRACE};
   $self->close_idle_connections;
   return $self;
+}
+
+sub _dbi_attr {
+  my ($self, $handle) = (shift, shift);
+  my $key = $self->dsn =~ m!^dbi:(\w+):! ? lc "$1_$_[0]" : "mysql_$_[0]";
+  return $handle->{$key} if @_ == 1;
+  $handle->{$key} = $_[1];
+  $handle;
 }
 
 sub _dequeue {
@@ -127,7 +127,7 @@ sub _dequeue {
   # especially once he discovers that DBD::mysql randomly reconnects under
   # you, silently, but only if certain env vars are set
   # hint: force-set mysql_auto_reconnect or whatever it's called to 0
-  $self->handle_attr($dbh, auto_reconnect => 0);
+  $self->_dbi_attr($dbh, auto_reconnect => 0);
 
   # Maintain Commits with Mojo::mysql::Transaction
   $dbh->{AutoCommit} = 1;
@@ -457,14 +457,6 @@ Parse configuration from connection string.
 
   # Username, database and additional options
   $mysql->from_string('mysql://batman@/db5?PrintError=1&RaiseError=0');
-
-=head2 handle_attr
-
-  $value = $self->handle_attr($dbh, "thread_id");
-  $value = $self->handle_attr($sth, "client_found_rows");
-  $self->handle_attr($sth, "client_found_rows");
-
-This method is EXPERIMENTAL and could be removed in future releases.
 
 =head2 new
 
--- a/lib/Mojo/mysql.pm
+++ b/lib/Mojo/mysql.pm
@@ -76,7 +76,7 @@ sub from_string {
   if (my $host = $url->host) { $dsn .= file_name_is_absolute($host) ? ";mysql_socket=$host" : ";host=$host" }
   if (my $port = $url->port) { $dsn .= ";port=$port" }
 
-  # Need to set the dsn before settings options
+  # Need to set the dsn before reading options
   $self->dsn($dsn);
 
   # Username and password
@@ -151,7 +151,7 @@ sub _set_strict_mode {
 
 =head1 NAME
 
-Mojo::mysql - Mojolicious and Async MySQL and MariaDB
+Mojo::mysql - Mojolicious and Async MySQL/MariaDB
 
 =head1 SYNOPSIS
 
@@ -165,7 +165,7 @@ Mojo::mysql - Mojolicious and Async MySQL and MariaDB
   # MySQL >= 8.0:
   my $mysql = Mojo::mysql->strict_mode('mysql://username:password@hostname/test;mysql_ssl=1');
 
-  # Connect to MariaDB instead of mysql
+  # Use DBD::MariaDB instead of DBD::mysql
   my $mysql = Mojo::mysql->strict_mode('mariadb://username@/test');
 
   # Create a table
@@ -234,9 +234,14 @@ Mojo::mysql - Mojolicious and Async MySQL and MariaDB
 
 =head1 DESCRIPTION
 
-L<Mojo::mysql> is a tiny wrapper around L<DBD::mysql> that makes
-L<MySQL|http://www.mysql.org> and L<MariaDB|https://mariadb.org/> a lot of fun
-to use with the L<Mojolicious|http://mojolicio.us> real-time web framework.
+L<Mojo::mysql> is a tiny wrapper around L<DBD::mysql> and L<DBD::MariaDB> that
+makes L<MySQL|http://www.mysql.org> and L<MariaDB|https://mariadb.org/> a lot
+of fun to use with the L<Mojolicious|http://mojolicio.us> real-time web
+framework.
+
+The two DBD drivers are compatible with both MySQL and MariaDB, but they offer
+different L</options>. L<DBD::MariaDB> should have better unicode support
+though and might become the default in the future.
 
 Database and handles are cached automatically, so they can be reused
 transparently to increase performance. And you can handle connection timeouts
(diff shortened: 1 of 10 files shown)


Grinnz:
To be clear: DBD::mysql and DBD::MariaDB both support both mysql and mariadb. The naming is unfortunate but DBD::MariaDB is just a fork with fixes including the fix to unicode handling which was briefly in DBD::mysql 4.042.

Grinnz:
Because of that I think it would be (further) confusing to indicate that one needs to use DBD::MariaDB to connect to MariaDB and vice versa.

jhthorsen:
@Grinnz: Thanks for the input!

The reason I'm holding this back is that I'm not sure if there are any drawbacks or noticeable slow downs in this PR.

I'm also unsure if it's a good idea to have DBD::MariaDB optional and DBD::mysql mandatory.

jhthorsen:
@Grinnz: Thumbs up or down on the PR?

Grinnz:
For my opinion: Mojo::mysql should just use DBD::MariaDB only for the same reason it was updated to require DBD::mysql 4.042. But there is the risk of someone relying on DBD::mysql's brokenness after the fix was reverted there.

jhthorsen:
That is interesting... I think maybe I could merge this with an additional warning if someone connects with "mysql://" and let the user know that DBD::MariaDB will be the default driver and then remove `handle_attr()` completely in the future.

Grinnz:
For the PR as is: the only issue I have is in the documentation, saying things like "# Connect to MariaDB instead of mysql" is inaccurate because either driver can (and often is) used for either database.

lib/Mojo/mysql.pm
```diff
@@ -453,16 +452,28 @@ Parse configuration from connection string.
...
   my $mysql = Mojo::mysql->new;
   my $mysql = Mojo::mysql->new(%attrs);
   my $mysql = Mojo::mysql->new(\%attrs);
   my $mysql = Mojo::mysql->new('mysql://user@/test');
+  my $mysql = Mojo::mysql->new('mariadb://user@/test');
 
 Construct a new L<Mojo::mysql> object either from L</ATTRIBUTES> and or parse
 connection string with L</"from_string"> if necessary.
 
+Connecting to MariaDB requires the optional module L<DBD::MariaDB> to be
```
Grinnz: And the other one being here.

Tekki:
That DBD::MariaDB starts with version 1 out of nothing shows a lot of self-confidence, but does anyone of you know how reliable this module is?
Making it more complicated: You can connect with DBD::mysql to all versions of MariaDB, but you only can connect to MySQL 8 if it was built with libmysqlclient-dev (and not libmariadbclient-dev). Did someone try DBD::MariaDB with MySQL 8?

Grinnz:
I can understand the skepticism without the context of the months of discussions and mostly silence on the list, but it is being maintained and supported by GoodData, far more actively than DBD::mysql has been in ten years; its 1.0 immediately solves my problems with unicode data; after working around DBD::mysql's fundamental unfixed issues forever that is enough for me.

Tekki:
Checked the repo now (should have done this before), there are hundreds of commits since 1.0, just they didn't yet publish a new version. If it solves unicode and other problems, we should have a way to support it.
~~How do we handle this in the tests if someone has installed DBD::MariaDB instead of DBD::mysql?~~

jhthorsen:
@Tekki: I've rewritten all the tests so they work with both drivers. You can try yourself with `TEST_ONLINE=mariadb://root@/test` or something.

Grinnz:
They just did a new release finally: https://metacpan.org/release/PALI/DBD-MariaDB-1.10 :tada:

jhthorsen:
I can't figure this out... I'm running the two commands below, but DBD::MariaDB complains a lot about:

> DBI::db=HASH(0x7fb219bace88)->disconnect invalidates 5 active statement handles (either destroy statement handles or call finish on them before disconnecting) at DBI.pm line 758.

```
TEST_PUBSUB=1 TEST_FOR=500 TEST_ONLINE=mysql://root@/test prove -lj4
TEST_PUBSUB=1 TEST_FOR=500 TEST_ONLINE=mariadb://root@/test prove -lj4
```

DBD::mysql on the other hand seems quite happy.

Versions installed:

* DBI 1.642
* DBD::mysql 4.049
* DBD::MariaDB 1.10 

Any ideas?

Do you know what this means @gooddata?

Tekki:
This was one of the next entries on my list. Still have some troubles to get the tests running.
I'm trying to prepare Docker images for the tests and I think we should test MySQL with DBD::mysql built with the libraries from MySQL and test MariaDB with DBD::MariaDB built with their libraries. And not mix them.
We still have DBD::mysql in Database.pm and in the cpanfile for example. When I correct this async_query_in_flight.t doesn't exit anymore.
I'll add the link to Dockerhub as soon as I think the image is usable (at the moment it is not).

Grinnz:
There's no reason to `use DBD::mysql;` as it will be automatically used by DBI when connecting.

Tekki:
[Here](https://hub.docker.com/r/tekki/mojolicious) is the Docker image. We need the tags `mariadb` and `mysql`. Copy the call to `prove` into a script, then:
```
docker container run --rm -v "$(pwd):/usr/src/app" tekki/mojolicious:mariadb ./mariadb_tests.pl
docker container run --rm -v "$(pwd):/usr/src/app" tekki/mojolicious:mysql ./mysql_tests.pl
```
But of course we get the same '...->disconnect invalidates...' warnings.
Attention: it is intended that the container processes the cpanfile on startup. I've commented this code block out for the moment, but it will be active again.

@Grinnz DBD::mysql is probably a regex replace from Mojo::Pg::Database.

@jhthorsen the problem mentioned above with the async tests was a simple error on my side (wrong IP address). As we have already seen with the failing connection to MySQL 8 these test don't produce errors, but hang forever if something doesn't work. Is there a chance that we can move them to the bottom, for example by renaming them to 'z_async...'?

jhthorsen:
@Tekki: The MySQL 8 issues sounds unrelated to this PR. Can you open a separate issue?

Tekki:
When do you plan to add this feature?
Don't forget that DBD::mysql is no longer a requirement (cpanfile).

jhthorsen:
The initial release will probably keep DBD::mysql as a requirement, because of https://github.com/jhthorsen/mojo-mysql/pull/47/files#diff-e06f03e15d5fd50a7f3927cd435d6e03R471

I can't merge this until I figure out https://github.com/jhthorsen/mojo-mysql/pull/47#issuecomment-448095164. Note however that it's not currently something I'm working actively on. Help is most certainly appreciated!

Tekki:
Asked for help at the Zurich Perl Meetup.
From people using Oracle I was told that they know the 'active statement handles' warning. It appears because DBI leaves open a handle as long as it expects that more data is going to read from it and in some cases this is too long.
I've looked around in the code and have no idea why we have this difference between the MySQL and the MariaDB module. For example DBD::mysql::st and DBD::MariaDB::st look the same. A very simple test where we get this warning is t/crud.t, and here DBD::MariaDB complains about 15 active handles.
Maybe we should ask @pali.

pali:
@Tekki do you have reproducer for this issue?

Tekki:
@pali The simplest way I found to reproduce it is the following:
```
#!/usr/bin/env perl
use Mojo::Base -strict;
use lib 'lib';
use Mojo::mysql;

my $mysql = Mojo::mysql->new($ENV{TEST_ONLINE});
my $db    = $mysql->db;
$db->query(
  'create table if not exists crud_test (
     id   serial primary key,
     name text
   )'
);

$db->insert('crud_test', {name => 'foo'});
my $result = $db->select('crud_test')->hashes->to_array;
END {$db and $db->query('drop table if exists crud_test');}
```
Taking the code from this PR,
```
$ export TEST_ONLINE='mysql://testuser:somepassword@somehost/test' && ./testscript.pl
```
produces no output, on the other hand:
```
$ export TEST_ONLINE='mariadb://testuser:somepassword@somehost/test' && ./testscript.pl
DBI::db=HASH(0x7ffff7c490d0)->disconnect invalidates 1 active statement handle...
```
The problem is with the `END` block.

pali:
Are you able to create a reproducer with only DBD::MariaDB module (i.e. without Mojo::Base and Mojo::mysql)? Because in your example are more layers and problem can be on any of them. And debugging all Mojo, DBI and DBD::MariaDB together is just hard.

jhthorsen:
Maybe you can modify/copy this test: https://github.com/jhthorsen/mojo-mysql/blob/master/t/test-dbi-async.t

Tekki:
@jhthorsen We don't need any async code, my script is the beginning and ending of crud.t without the test commands and even here we get the warning. Pali asks for an example that directly uses DBD::MariaDB.

janchochol:
I think, that change in behavior of ``DBD::MariaDB`` was introduced in https://github.com/gooddata/DBD-MariaDB/commit/ea8685409ac67f4c3fe235f38b53c39fcfd68e0c.
It can be fixed on mojo side by:
```diff
diff --git a/lib/Mojo/mysql/Results.pm b/lib/Mojo/mysql/Results.pm
index 5fbaaaf..e3bfc5f 100644
--- a/lib/Mojo/mysql/Results.pm
+++ b/lib/Mojo/mysql/Results.pm
@@ -75,6 +75,7 @@ sub _expand {
 sub DESTROY {
   my $self = shift;
   return unless my $db = $self->{db} and my $sth = $self->{sth};
+  $sth->finish;
   push @{$db->{done_sth}}, $sth;
 }
```

But according to ``DBI`` help documentation:
> When all the data has been fetched from a SELECT statement, the driver will automatically call finish for you. So you should not call it explicitly except when you know that you've not fetched all the data from a statement handle and the handle won't be destroyed soon.

So the change is problematic, but without it, it is not possible to get attributes in some case (it is too soon before ``fetchall*``, but too late after ``fetchall*``).

We must think about correct fix (cc @pali).

Tekki:
This change in Results.pm doesn't solve all the problems, instead there are new warnings about '...results for the wrong handle...' in database.t. Even if we surround it with `eval { $sth->finish; };` there is one remaining open handle in database.t, pubsub.t and results_methods.t.

janchochol:
That is only temporary workaround (sorry for not being clear) - we must fix it ``DBD::MariaDB`` as it goes against ``DBI`` specification.

Tekki:
I've understood it this way. Just jhthorsen will not merge this PR as long as we get any warnings. So I think we will wait until DBD::MariaDB gets updated.

pali:
@Tekki: I looked at it and I have prepared some fixes. But still I would like to include test for this problem to mimics mojo-mysql behavior.

pali:
I created pull request https://github.com/gooddata/DBD-MariaDB/pull/129 for DBD::MariaDB with a proposed fix for this problem. Still I would like to see a test which uses just DBD::MariaDB and similate problem described in this issue. Otherwise there is no proof that problem is really in DBD::MariaDB and not in combination DBD::mysql with mojo-mysql.

Tekki:
@pali, try this one:
```
#! /usr/bin/env perl
use strict;
use warnings;

use DBI;

my $dbh = DBI->connect(split /--/, $ENV{TEST_DB});
  
my $sth = $dbh->prepare('select * from mojo_json_test');
$sth->execute or die $dbh->errstr;
print $sth->rows. " rows found.\n";
while (my $ref = $sth->fetchrow_hashref) {
    print "Found a row: id = $ref->{'id'}, name = $ref->{'name'}\n";
}

END {
  $sth->finish;
}
```
Warns with 'DBD::MariaDB disconnect_all: 1 database handlers were not released'.

pali:
@Tekki Pull request https://github.com/gooddata/DBD-MariaDB/pull/129 fixes problem that `$sth->finish` is not needed to call after fetching all rows (i.e. when `$sth->fetchrow_hashref` returns undef then $sth->fetchrow_hashref is not needed to call).

But your above example discovered two new and totally different problems:

1) If you remove `die` and `mojo_json_test` table does not exist then `$sth->rows` returns 18446744073709551615 instead of -1 (unknown number of rows / error)

2) When `$sth` is referenced in `END` block then DBI or Perl first destroy `$dbh` and later `$sth`. And when DBI calls DBD::MariaDB's `disconnect_all` then for unknown reasons driver's `ChildHandles` is empty even when there some `$dbh`s active...

First one is bug in DBD::MariaDB which should be easy to fix. Second one seems to be a problem in DBI (or maybe Perl itself) and seems it is DBI driver independent.

pali:
As a workaround for 2) you can call `$dbh->disconnect` after `$sth->finish`.

Tekki:
I've written this script because I tried to reproduce the warning in a simple way, but I don't know enough about the internals of Mojo::mysql to really find its cause. There is no warning with a DBI::mysql URL in $ENV{TEST_DB}, so I'm not sure about your point 2).
The 184... in 1) looks like a mismatch between signed and unsigned integers:
18446744073709551615 - 2^64 = -1

pali:
> There is no warning with a DBI::mysql URL in $ENV{TEST_DB}, so I'm not sure about your point 2).

That is because DBD::mysql does not implement `disconnect_all` method. So it never throw any such warning.

> The 184... in 1) looks like a mismatch between signed and unsigned integers:
18446744073709551615 - 2^64 = -1

Yes, mysql server returns 64bit unsigned integer and maximal value means error.

pali:
Fix for `$sth->rows` is there: https://github.com/gooddata/DBD-MariaDB/pull/130

pali:
Fix for `disconnect_all` is there: https://github.com/gooddata/DBD-MariaDB/pull/131

It is just fix for DBD::MariaDB driver. Instead of `ChildHandles`, DBD::MariaDB now tracks own list of child handles in own double linked list C structure.

In future, please report DBD::MariaDB related bugs to our bug tracker:
https://github.com/gooddata/DBD-MariaDB/issues

choroba:
Can anyone verify the fix in DBD::MariaDB solved the issue? Thanks.

Tekki:
@pali @choroba: No more warnings when I run the tests with the most recent version of DBD::MariaDB from Github.
But the following doesn't work:
```
use Mojo::Base -strict;
use Test::More;

plan skip_all => 'TEST_ONLINE=mysql://root@/test' unless $ENV{TEST_ONLINE};

use Mojo::mysql;

my $mysql = Mojo::mysql->new($ENV{TEST_ONLINE});

note 'One';
my $db = $mysql->db;
is 1, 1;

note 'Two';
$db->disconnect;
is 2, 2;

note 'Three';
$db = $mysql->db;
is 3, 3;

done_testing;
```
Returns:
```
# One
ok 1
# Two
ok 2
# Three
All 2 subtests passed
```
Seems I have to open an issue on DBD::MariaDB.

choroba:
Please do. If you could reproduce the issue without the Mojo wrappings it would be the best :-)

Tekki:
[Here](https://github.com/gooddata/DBD-MariaDB/issues/137) is the issue on DBD::MariaDB.

Tekki:
@jhthorsen the developers at Gooddata found out that this is an issue with libmysqlclient-dev.

Tekki:
With DBD::MariaDB 1.21 everything looks fine on different versions of MySQL and MariaDB. Tested with libmysqlclient-dev 8.0.15.

jhthorsen:
I think I've covered all the issues mentioned in this PR. If not, then please open new issues.

Thanks for all the help to everyone involved!

jhthorsen:
I think I've covered all the issues mentioned in this PR. If not, then please open new issues.

Thanks for all the help to everyone involved!

PR #48, opened 2018-12-09
Initialize SSL connection to MySQL 8

Tekki:
Connections to MySQL 8 are a bit more complicated/sophisticated. The idea of the new caching_sha2_password authentication method is that we only need the overhead of SSL for the first connection of a user. This is used in `t/01-init.t`, so tests on MySQL 8 can be run without additional parameters. Maybe somewhere in the future DBD::mysql will handle this automatically.

### DIFF
--- a/lib/Mojo/mysql.pm
+++ b/lib/Mojo/mysql.pm
@@ -146,7 +146,7 @@ Mojo::mysql - Mojolicious and Async MySQL
 
   # Connect to a remote database
   my $mysql = Mojo::mysql->strict_mode('mysql://username:password@hostname/test');
-  # MySQL >= 8.0:
+  # MySQL >= 8.0, initial connection for username:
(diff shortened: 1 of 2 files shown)


Tekki:
The reason why I created this new script is that the first tests on the db are `t/async_query_in_flight.t`. If the connection to MySQL 8 can't be established they hang forever.

jhthorsen:
I don't understand what value this PR adds.

Tekki:
Re-thinking it, I don't understand it either.
I had the idea that in this way it is possible to run the tests on 8.0 without having to care about the caching_sha2_password stuff. The whole thing is not documented well. But on the other hand it is highly hypothetical that someone installs 8.0, builds DBD::mysql in a way that it is able to connect to it and at the end doesn't know what he is doing.
Feel free to close the PR.

PR #50, opened 2018-12-19
Improve error handling in async query tests

Tekki:
If the connection to the database fails (wrong IP address, missing encryption) async_query_in_flight.t doesn't produce an error, but hangs forever. If we move it to the bottom we get an error from an earlier test.


jhthorsen:
I’m not going to take this PR. If an async test does not finish, we should force the test to fail with a “master timer”. Something like this:

```
Mojo::IOLoop->timer(10 => sub { Mojo::IOLoop->stop });
```

Tekki:
I see, to move the problems out of the way wasn't one of my brightest ideas.
Let's take a look at your solution. If we force the IOLoop to stop, we get the following message:

>'sqqqqqqqqqq' doesn't match '(?^:^s\w{200}$)'

It's not much more helpful than a hanging process, isn't it? Wouldn't it be nice if we get something like the following instead?

>DBI connect('dbname=test;mysql_ssl=1;host=127.0.0.1;port=33080','testuser',...) failed: Can't connect to MySQL server on '127.0.0.1'

My interpretation is that the tests don't exit because `$mysql->db` fails and we never arrive at `$delay->begin`. If this is true, we could for example test `$mysql->db` before we start the IOLoop.

jhthorsen:
How about appending `->catch(sub { $err = pop; Mojo::IOLoop->stop })` to the `delay()` call instead? Should‘n't that catch the failing `db()` call inside the step?

Tekki:
If you push the error to `@err` it works (`$err` is not declared at this place). I just wonder if a construction like
```
    Mojo::IOLoop->delay( ... )->catch( sub { push @err, pop; ... } );
```
is considered as best practice.

jhthorsen:
It’s the way to do it

Tekki:
So it would be nice if it's mentioned in the docs for [Mojo::IOLoop](https://github.com/mojolicious/mojo/blob/master/lib/Mojo/IOLoop.pm#L404).
```
my $delay = $loop->delay(sub {...})->catch(sub {...});
```
Ping @kraih.

jhthorsen:
It does:

> Mojo::IOLoop::Delay inherits all methods from Mojo::Promise and implements the following new ones.

jhthorsen:
Thanks!

PR #51, opened 2019-01-14
Extended JSON support

Tekki:
As discussed in https://github.com/Tekki/mojo-mysql/issues/1

### DIFF
--- a/lib/SQL/Abstract/mysql.pm
+++ b/lib/SQL/Abstract/mysql.pm
@@ -1,7 +1,19 @@
 package SQL::Abstract::mysql;
 use Mojo::Base 'SQL::Abstract';
 
+use Mojo::JSON 'encode_json';
+
 BEGIN { *puke = \&SQL::Abstract::puke }
+
+sub new {
+  my $self = shift->SUPER::new(@_);
+
+  # -e and -ne op
+  push @{$self->{unary_ops}}, {regex => qr/^e$/, handler => '_where_op_EXISTS',},
+    {regex => qr/^ne$/, handler => '_where_op_NOT_EXISTS',};
+
+  return $self;
+}
 
 sub insert {
   my $self    = shift;
@@ -33,6 +45,36 @@ sub insert {
   $sql = join ' ', $self->_sqlcase("$command into"), $table, $sql;
 
   return wantarray ? ($sql, @bind) : $sql;
+}
+
+sub _insert_value {
+  my ($self, $column, $v) = @_;
+
+  if (ref $v eq 'HASH') {
+
+    # THINK: anything useful to do with a HASHREF ? (SQL::Abstract)
+    # ANSWER: Of course, insert JSON!
+    $v = encode_json($v);
+  }
+  return $self->SUPER::_insert_value($column, $v);
+}
+
+sub _json_extract {
+  my ($self, $label, $alias) = @_;
+  return $self->SUPER::_quote($label) unless $label =~ /->/;
+
+  my ($field, $unquote, $path) = $label =~ /(.+)->([>]?)(.+)/;
+  $field = $self->SUPER::_quote($field);
+
+  my $rv = "JSON_EXTRACT($field,'\$.$path')";
+  $rv = "JSON_UNQUOTE($rv)" if $unquote;
+
+  if ($alias) {
+    $path =~ /(\w+)$/;
+    $rv .= " AS $self->{quote_char}$1$self->{quote_char}";
+  }
+
+  return $rv;
 }
 
 sub _order_by {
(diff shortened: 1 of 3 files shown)


jhthorsen:
Hey! I think this is a great extension, but I do have some questions: Is any JSON document supported? Like what about strings (not sure if that makes sense), but at least ARRAY on the top level? I wonder if it would be safer to do: `{column => {json => $whatever}}`. I know it's a bit more verbose, but at least it will always work.

The whole thing about inserting documents is a bit confusing to me, since I thought that was already handled by Mojo::mysql::Database.

I also see some commas that is not required, like `,}`, but I can clean that up later on...

I'll have a closer look tomorrow, but this will get merged on way or another 👍

Tekki:
We look at the things from a different point of view.
- `,` at the end of lists is a question of style and the style in this repo is defined by you
- think of a document as a complete row and not just as values of one column; first dimension of keys are the columns, deeper dimensions are documents inside columns
- forget about 'JSON'; how documents are implemented in a database is a technical question that a user of an abstract module shouldn't be confronted with
- the abstract module should work standalone and not just in conjunction with Database
- there are no tests for inserting an array at the first level, so there's a chance it doesn't work yet

lib/SQL/Abstract/mysql.pm
```diff
@@ -35,6 +47,36 @@ sub insert {
   return wantarray ? ($sql, @bind) : $sql;
 }
 
+sub _insert_value {
+  my ($self, $column, $v) = @_;
+
+  if (ref $v eq 'HASH') {
```
jhthorsen: Is there a downside to supporting the same syntax as in https://github.com/jhthorsen/mojo-mysql/blob/38c16bdc6822e2039a258b46170e3874541fa921/lib/Mojo/mysql/Database.pm#L311

Like this:

```
if (ref $v eq 'HASH' and exists $param->{json}) { ... }
```

It's a bit extra to type, but it would allow it to be extended further in the future.

lib/SQL/Abstract/mysql.pm
```diff
@@ -1,8 +1,20 @@
...
+use Mojo::JSON 'encode_json';
+
 BEGIN { *puke = \&SQL::Abstract::puke }
 
+sub new {
+  my $self = shift->SUPER::new(@_);
+
+  # -e and -ne op
+  push @{$self->{unary_ops}}, {regex => qr/^e$/, handler => '_where_op_EXISTS',},
+    {regex => qr/^ne$/, handler => '_where_op_NOT_EXISTS',};
```
jhthorsen: If you want to be "git blamed" for these lines, then you have to remove the extra comma before `}`.

lib/SQL/Abstract/mysql.pm
```diff
@@ -98,6 +140,95 @@ sub _order_by {
...
 }
 
+sub _order_by_chunks {
+  my ($self, $arg) = @_;
+  if ($arg =~ /->/) {
+    return $self->_json_extract($arg);
+  }
+  else {
+    return $self->SUPER::_order_by_chunks($arg);
+  }
```
jhthorsen: Could you compress this to:

```
sub _order_by_chunks {
  my ($self, $arg) = @_;
  return $self->_json_extract($arg) if $arg =~ /->/;
  return $self->SUPER::_order_by_chunks($arg);
}
```

Same with `_select_fields()` below.

lib/SQL/Abstract/mysql.pm
```diff
@@ -98,6 +140,95 @@ sub _order_by {
...
+}
+
+sub _update_set_values {
+  my ($self, $data) = @_;
+
+  for my $k (sort grep /->/, keys %$data) {
+    my ($label, $path) = $k =~ /(.+)->(.*)/;
+    my $origin = $self->_quote($label);
+    puke "you can\'t update $label and its values in the same query"
+      unless ($data->{$label}->[0] || 'json') =~ /^json/i;
```
jhthorsen: I'm not a fan of the extra `->`. Please write it as `$data->{$label}[0]`. (Seen other places in this PR as well)

lib/SQL/Abstract/mysql.pm
```diff
@@ -98,6 +140,95 @@ sub _order_by {
...
+  else {
+    return join ',', map { $self->_quote($_) } @$fields;
+  }
+}
+
+sub _update_set_values {
+  my ($self, $data) = @_;
+
+  for my $k (sort grep /->/, keys %$data) {
+    my ($label, $path) = $k =~ /(.+)->(.*)/;
```
jhthorsen, line 158: Is there a test for when this regex fails to match `$k`? I'm unsure from reading the code about what happens if `$k` is simply "foo", meaning without `->`.

lib/SQL/Abstract/mysql.pm
```diff
@@ -98,6 +140,95 @@ sub _order_by {
...
+
+  if (grep /->/, @$fields) {
+    return join ',', map { $self->_json_extract($_, 1) } @$fields;
+  }
+  else {
+    return join ',', map { $self->_quote($_) } @$fields;
+  }
+}
+
+sub _update_set_values {
```
jhthorsen, line 153: I think this whole method is a bit hard to read, because of all the nested if/else. Can you move some of the code out in methods with descriptive names? I think that could limit the nesting by using `return` instead.

lib/SQL/Abstract/mysql.pm
```diff
@@ -98,6 +140,95 @@ sub _order_by {
...
+    if (defined $data->{$k}) {
+      if ($data->{$label}->[0]) {
+        puke "you can\'t update and remove values of $label in the same query"
+          unless $data->{$label}->[0] =~ /^json_set/i;
+      }
+      else {
+        $data->{$label}->[0] = $path ? $self->_sqlcase('json_set(') . "$origin)" : '?';
+      }
+      my $placeholder;
+      if (ref $data->{$k}) {
```
jhthorsen, line 174: Is this symmetric with what you have at https://github.com/jhthorsen/mojo-mysql/pull/51/files#diff-45bdaf6ecc6b539751f37122c23c1f17R53 ?

`if (ref $v eq 'HASH') { ...`

lib/SQL/Abstract/mysql.pm
```diff
@@ -98,6 +140,95 @@ sub _order_by {
...
+}
+
+sub _where_hashpair_SCALAR {
+  my ($self, $k, $v) = @_;
+
+  if ($k =~ /->/) {
+    $k = \$self->_json_extract($k);
+  }
+
+  return $self->SUPER::_where_hashpair_SCALAR($k, $v);
```
jhthorsen, line 214: This method could be compressed to something like:

```
$k = \$self->_json_extract($k) if $k =~ /->/;
return $self->SUPER::_where_hashpair_SCALAR($k, $v);
```

jhthorsen (changes requested):
Sorry for all the nitpicking, but there are some parts of this code that I have a hard time wrapping my head around. However, the documentation seems good and there's a bunch of new tests, so I'm still very much positive.

Another thing: What about `$obj->TO_JSON`? I think that should be supported as well.

Maybe @shadowcat-mst has some input on the syntax? Especially regarding `{column => \%json}` v.s `{column => {json => \%json}}`.

Tekki:
I've no problems with your nitpicking and there is no hurry for this new feature.
In general I'm not too happy about the way I've implemented the changes. SQL::Abstract gives us a lot of possibilities to override internal methods and I tried not to disable too much of the original code. But unfortunately most of the time this possibility is missing, so I had to use maybe 3 different approaches.

If you still ask about `column => {json => $whatever}` it is because you didn't forget about JSON. I give you an example: Let's say we have a table with columns `firstname`, `lastname` and `address`. From the API you get a POST request with
```
$request = {
  firstname => 'Haymitch',
  lastname => 'Abernathy',
  address => {place => 'Victors Village', district => 12}
}
```
Now what to expect from an abstract module is that it knows how to handle `->insert($table, $request)` so you don't have to split it into
```
{
  firstname => $request->{firstname},
  lastname => $request->{lastname},
  address => {json => $request->{address}}
}
```
As I said earlier, the user should't have to worry about the technical way how the database stores a hash reference.

lib/SQL/Abstract/mysql.pm
Tekki: A shorter answer than the one below: We don't write
`$db->query('insert into foo values (?)', {integer => 23});` either.

lib/SQL/Abstract/mysql.pm
Tekki: If the key is "foo" it will not appear in `$k` because of the grep. This block only processes keys with arrows.

Tekki:
Tried to discuss this topic too at the Zurich Perl Meetup. But the discussion immediately went to the question why we should use JSON at all. We were looking for real use cases and until we found one (postal addresses) the time was over.

lib/SQL/Abstract/mysql.pm
Tekki: It's not symmetric and `if (ref $v eq 'HASH') { ...` is wrong because it doesn't accept array references (the top level of a document can be an array, as you have mentioned in your first comment).
At this place we process the value of `field->some.key` and we can expect that everything that is a reference has to be JSON encoded.

lib/SQL/Abstract/mysql.pm
Tekki: It's hard to read, but there are too many variables for a method and the block is only 40 lines long. Maybe some comments will help?

jhthorsen:
@Tekki: I think this is a great enhancement and it also resolves #57. Any way you can resolve the conflict and (force) push other cleanups?

If you want, we can mark this feature as EXPERIMENTAL, even though I don't think we'll get rid of it once it's merged. I do like "EXPERIMENTAL" though, in case there's parts that need tweaking or if there is a big chance for bugs.

Tekki:
I have to take a look at it again, since we made several changes in the meantime.

jhthorsen:
I'm closing this for now, since it seems to have gone stale. Please do create a new branch in this repo if you want to complete it, now that you have commit bit.

Tekki:
It has gone stale because we had other things to do. But the code has changed a lot in the meantime and this needs a new approach, so it's okay to close it.

Issue #52, opened 2019-01-31
Unstable expand function


Tekki:
`$db->query(...)->expand->something` crashes under several circumstances. I'll prepare a PR with failing tests.

jhthorsen:
Feel free to close this, since you created #53.

jhthorsen:
I'm closing this, since #58 is so close.

Excellent work @Tekki!

PR #53, opened 2019-02-01
Improve Result->expand

Tekki:
Here is a first commit for #52 with some more tests.
Line 34 and 36 are ok.
Line 38 and 42: `Not a HASH reference at .../Results.pm line 52.`
Line 32 and 40, this is the interesting one:
`DBD::mysql::st FETCH failed: statement contains no result at .../Results.pm line 59.`
The circle is closing with #47 and https://github.com/gooddata/DBD-MariaDB/pull/24, isn't it?

### DIFF
--- a/t/json.t
+++ b/t/json.t
@@ -20,12 +20,26 @@ is $db->query('select json_extract(j, "$.foo") from mojo_json_test')->array->[0]
 is_deeply $db->query('select id, name, j from mojo_json_test where json_extract(j, "$.foo") = 42')->expand->hash,
   {id => $$, name => $0, j => {foo => 42}}, 'expand json';
 
+my $value_hash = {nick => 'supergirl'};
+my $value_json = Mojo::JSON::to_json($value_hash);
+my $query      = 'select name from mojo_json_test where name like "%supergirl%"';
 $db->query('insert into mojo_json_test (name) values (?)', {json => {nick => 'supergirl'}});
-is_deeply $db->query('select name from mojo_json_test where name like "%supergirl%"')->expand->hash,
-  {name => Mojo::JSON::to_json({nick => 'supergirl'})}, 'name as string';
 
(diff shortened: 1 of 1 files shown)


Tekki:
I wonder why the Travis tests pass. I'm on Perl 5.28.1 with DBI 1.642 and DBD::mysql 4.050.

jhthorsen:
@Tekki: Which version of MySQL?

Tekki:
I was on 5.7. Now I've tested it on 8.0 and MariaDB 10.3 and 10.4 too, with the same result. "Not a HASH reference" is an error inside Results. It looks as if those tests are not executed in Travis.

Btw. I didn't find `$sth->{mysql_type}` from [l. 59](https://github.com/jhthorsen/mojo-mysql/blob/master/lib/Mojo/mysql/Results.pm#L59) in the docs. Is this equal to `$sth->{TYPE}` in [DBI](https://metacpan.org/pod/DBI#TYPE)?

pali:
Attributes which start with lowercase character and with DBI driver prefix are DBI driver specific, e.g. DBD::mysql has driver specific prefix `mysql_` and therefore `$sth->{mysql_type}` is non-portable DBI driver specific attribute. It is documented in https://metacpan.org/pod/DBD::mysql#mysql_type and because it is driver specific it would not work with any other DBI driver (including DBD::MariaDB). Same applies for `$sth->{mysql_insertid}` or `$sth->{mysql_warning_count}`.

pali:
And...
> Line 32 and 40, this is the interesting one:
> `DBD::mysql::st FETCH failed: statement contains no result at .../Results.pm line 59.`

Pull request https://github.com/gooddata/DBD-MariaDB/pull/24 for DBD::MariaDB fixed exactly above problem.

Tekki:
Yes, that's why I'm speaking about the closing circle. I think we can solve this here by getting this value earlier.
Thanks for the clarification about `mysql_type`.

Tekki:
I'll prepare a new PR based on the actual code.

Issue #54, opened 2019-02-18
SQL::Abstract::mysql join does not match documentation


daleif:
I'm playing with the `join` examples from the SQL::Abstract::mysql` POD

For example

    my $mysql = Mojo::mysql->new;
    my $abstract = $mysql->abstract;
    # "select * from foo left join bar on (bar.foo_id = foo.id)"
    say $abstract->select(['foo', [-left => 'bar', foo_id => 'id']]);

just gives

    SELECT * FROM `foo`, `ARRAY(0x1c53010)`

Am I missing something really obvious here?

I've tested some of the other nested array ref examples, neither seem to work. Also, not that it does not seem that `t/sql.t` tests any nested array ref examples.

jhthorsen:
I don’t think I will find time to look at this until later this week, but maybe @Tekki knows?

Tekki:
I'm the one who knows nothing, but it seems I forgot to steal a part of the code from Mojo::Pg. I'll take a look at it (@jhthorsen could care about DBD::MariaDB in the meantime).

Tekki:
Made some additions to the module in #56. Thanks to @daleif for reporting this issue.

jhthorsen:
This should be fixed by @Tekki. Hopefully I’ll be able to make a new release later today.

PR #56, opened 2019-02-19
Add missing code for SQL JOIN

Tekki:
Changes for #54. Now this should be the correct PR.


jhthorsen (changes requested):
I think this is a great improvement, but I don’t want do/while loops.

Also, did you run the code through perltidy?

Tekki:
Yes, I ran perltidy. And I don't want do...while loops either.

Tekki:
Does it look better now?

jhthorsen (approved):
Yes. Looks good :)

Issue #57, opened 2019-03-01
expand mangles unicode characters stored in JSON column


ViktorNacht:
I can create a working example, but just a quick heads-up that I'm getting mangled unicode when I use the expand function selecting data from a JSON column (technically I was using expand(1)). 

I confirmed this by removing expand and then de-serializing it with j and the unicode characters remained in-tact.

FYI, my DSN is something like:

mysql://foo:bar@db.us-west-2.rds.amazonaws.com/table_name?mysql_enable_utf8mb4=1

And I don't know if it matters, but I'm inserting the data by serializing it with j first. I haven't figured out how to use json with SQL::Abstract::mysql

Grinnz:
Serializing with j is incorrect. Since the query and parameters are being encoded to UTF-8 already, you want to use to_json so as not to double-encode. Alternatively, use j or encode_json and bind it as a SQL_BLOB as shown in https://metacpan.org/pod/Mojo::mysql::Database#query. But you might as well just use the json generator.

ViktorNacht:
Cool. Not to disagree with you, but to make sure I understand what's happening.

I'm using $db->insert() because passing a hash reference with the column names and values is better for my application. Because I didn't see a way to use the json generator without using $db->query, I encoded the json column with j, and then inserted with $db->insert(). 

Selecting the json column with the mysql CLI and via a graphical interface both show the ñ character correctly, so it *appears* it was not double-encoded. 

Only when I use ->expand does it get mangled. Using $db->select() and j to deserialize returns the proper character. 

What I am missing I this process? I'm just curious how using j in and j out works fine, but mixing j with ->expand doesn't.

Thank you for your time, I appreciate the learning process.

Grinnz:
It is hard to debug encoding issues because the tools you use to debug also work in a certain encoding. When you select (with utf8mb4 enabled) and then use j to deserialize, both of those steps decoded, so it successfully double-decodes the double encoded data.

Grinnz:
You are correct that you can't use the json generator with the SQL::Abstract functions. SQL::Abstract::Pg added this feature (https://metacpan.org/pod/SQL::Abstract::Pg#JSON) but it hasn't been added to SQL::Abstract or SQL::Abstract::mysql yet.

jhthorsen:
@ViktorNacht: You might want to have a look at #51. I'll merge the changes once @Tekki is happy with the implementation. Any feedback would be very much appreciated, so please take the PR for a spin.

[...]

PR #58, opened 2019-03-06
Correct handling of fetchall and arrays in Result

Tekki:
Changes for #52.

### DIFF
--- a/lib/Mojo/mysql/Results.pm
+++ b/lib/Mojo/mysql/Results.pm
@@ -7,9 +7,9 @@ use Mojo::Util 'tablify';
 
 has [qw(db sth)];
 
-sub array { ($_[0]->_expand({}))[0] }
+sub array { ($_[0]->_expand({list => 0, type => 'array'}))[0] }
 
-sub arrays { _c($_[0]->_expand({list => 1})) }
+sub arrays { _c($_[0]->_expand({list => 1, type => 'array'})) }
 
 sub columns { shift->sth->{NAME} }
 
@@ -17,9 +17,9 @@ sub expand { $_[0]{expand} = defined $_[1] ? 2 : 1 and return $_[0] }
 
 sub finish { shift->sth->finish }
 
-sub hash { ($_[0]->_expand({hash => 1}))[0] }
+sub hash { ($_[0]->_expand({list => 0, type => 'hash'}))[0] }
 
-sub hashes { _c($_[0]->_expand({hash => 1, list => 1})) }
+sub hashes { _c($_[0]->_expand({list => 1, type => 'hash'})) }
 
 sub rows { shift->sth->rows }
 
@@ -49,36 +49,40 @@ sub _expand {
   my ($idx, $names) = $mode == 1 ? $self->_types : ();
 
   # Fetch sql data
-  my $sql_data;
-  if ($to->{list}) {
-    $sql_data = $to->{hash} ? $self->sth->fetchall_arrayref({}) : $self->sth->fetchall_arrayref;
-  }
-  else {
-    @$sql_data = $to->{hash} ? $self->sth->fetchrow_hashref : $self->sth->fetchrow_arrayref;
-  }
+  my $hash = $to->{type} eq 'hash';
+  my $sql_data
+    = $to->{list} && $hash ? $self->sth->fetchall_arrayref({})
+    : $to->{list} ? $self->sth->fetchall_arrayref
+    : $hash       ? [$self->sth->fetchrow_hashref]
+    :               [$self->sth->fetchrow_arrayref];
 
-  # expand(), only expand json columns
-  if ($mode == 1) {
-    if ($to->{hash}) {
-      for my $r (@$sql_data) {
-        $r->{$_} = from_json $r->{$_} for grep { defined $r->{$_} } @$names;
-      }
-    }
-    else {
-      for my $r (@$sql_data) {
-        $r->[$_] = from_json $r->[$_] for grep { defined $r->[$_] } @$idx;
-      }
-    }
-  }
-
-  # expand(1), force expanding
-  elsif ($mode == 2) {
-    for my $r (@$sql_data) {
-      $_ = from_json $_ for grep /^[\[{].*[}\]]$/, $to->{hash} ? values %$r : @$r;
-    }
+  # Optionally expand
+  if ($mode) {
+    my $from_json = __PACKAGE__->can(sprintf '_from_json_mode_%s_%s', $mode, $to->{type});
+    $from_json->($_, $idx, $names) for @$sql_data;
   }
 
   return @$sql_data;
+}
+
+sub _from_json_mode_1_array {
+  my ($r, $idx, $names) = @_;
+  $r->[$_] = from_json $r->[$_] for grep { defined $r->[$_] } @$idx;
+}
+
+sub _from_json_mode_1_hash {
+  my ($r, $idx, $names) = @_;
+  $r->{$_} = from_json $r->{$_} for grep { defined $r->{$_} } @$names;
+}
+
+sub _from_json_mode_2_array {
+  my ($r, $idx, $names) = @_;
+  $_ = from_json $_ for grep /^[\[{].*[}\]]$/, @$r;
+}
+
+sub _from_json_mode_2_hash {
+  my ($r, $idx, $names) = @_;
+  $_ = from_json $_ for grep /^[\[{].*[}\]]$/, values %$r;
 }
 
 sub _types {
@@ -86,7 +90,7 @@ sub _types {
   return @$self{qw(idx names)} if $self->{idx};
 
   my $types = $self->db->mysql->_dbi_attr($self->sth, 'type');
-  my @idx   = grep { $types->[$_] == 245 or $types->[$_] == 252 } 0 .. $#$types;    # 245 = MySQL, 252 = MariaDB
+  my @idx = grep { $types->[$_] == 245 or $types->[$_] == 252 } 0 .. $#$types;    # 245 = MySQL, 252 = MariaDB
 
   return ($self->{idx} = \@idx, $self->{names} = [@{$self->columns}[@idx]]);
 }
--- a/lib/Mojo/mysql/Results.pm
+++ b/lib/Mojo/mysql/Results.pm
@@ -48,21 +48,11 @@ sub _expand {
 
   # Force expanding
   if ($mode == 2) {
-    my $is_json = qr/^[\[{].*[}\]]$/;
-    if (ref $rows[0] eq 'HASH') {
-      return map {    ## no critic
-        my $r = $_;
-        $_ = from_json $_ for grep {$is_json} values %$r;
-        $r;
-      } @rows;
-    }
-    else {
-      return map {    ## no critic
-        my $r = $_;
-        $_ = from_json $_ for grep {$is_json} @$r;
-        $r;
-      } @rows;
-    }
+    my $looks_like_json = qr/^[\[{].*[}\]]$/;
+    return map {    ## no critic
+      $_ = from_json $_ for grep {$looks_like_json} ref eq 'HASH' ? values %$_ : @$_;
+      $_;
+    } @rows;
   }
 
   # Only expand json columns
--- a/lib/Mojo/mysql/Results.pm
+++ b/lib/Mojo/mysql/Results.pm
@@ -5,8 +5,6 @@ use Mojo::Collection;
 use Mojo::JSON 'from_json';
 use Mojo::Util 'tablify';
 
-my $LOOKS_LIKE_JSON_RE = qr/^[\[{].*[}\]]$/;
-
 has [qw(db sth)];
 
 sub array { ($_[0]->_expand({}))[0] }
@@ -76,7 +74,7 @@ sub _expand {
   # expand(1), force expanding
   elsif ($mode == 2) {
     for my $r (@$sql_data) {
-      $_ = from_json $_ for grep $LOOKS_LIKE_JSON_RE, $to->{hash} ? values %$r : @$r;
+      $_ = from_json $_ for grep /^[\[{].*[}\]]$/, $to->{hash} ? values %$r : @$r;
     }
   }
 
--- a/lib/Mojo/mysql/Results.pm
+++ b/lib/Mojo/mysql/Results.pm
@@ -5,11 +5,13 @@ use Mojo::Collection;
 use Mojo::JSON 'from_json';
 use Mojo::Util 'tablify';
 
+my $LOOKS_LIKE_JSON_RE = qr/^[\[{].*[}\]]$/;
+
 has [qw(db sth)];
 
-sub array { ($_[0]->_expand($_[0]->sth->fetchrow_arrayref))[0] }
+sub array { ($_[0]->_expand({}))[0] }
 
-sub arrays { _c($_[0]->_types->_expand(@{$_[0]->sth->fetchall_arrayref})) }
+sub arrays { _c($_[0]->_expand({list => 1})) }
 
 sub columns { shift->sth->{NAME} }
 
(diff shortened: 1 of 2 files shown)


lib/Mojo/mysql/Results.pm
```diff
@@ -9,7 +9,7 @@ has [qw(db sth)];
 
 sub array { ($_[0]->_expand($_[0]->sth->fetchrow_arrayref))[0] }
 
-sub arrays { _c($_[0]->_expand(@{$_[0]->sth->fetchall_arrayref})) }
+sub arrays { _c($_[0]->_types->_expand(@{$_[0]->sth->fetchall_arrayref})) }
```
jhthorsen: I don't think we should figure out the types, unless we're going to actually expand.

lib/Mojo/mysql/Results.pm
```diff
@@ -47,19 +47,27 @@ sub _expand {
...
+        $r;
+      } @rows;
+    }
+    else {
+      return map {    ## no critic
+        my $r = $_;
+        $_ = from_json $_ for grep {$is_json} @$r;
+        $r;
+      } @rows;
+    }
```
jhthorsen: This looks like too many lines. Can it be written like this? (And maybe rename `$is_json` to `$json_re` or `$looks_like_json`?)

```
if ($mode == 2) {
  return map {
    $_ = from_json $_ for grep {$is_json} ref eq 'HASH' ? values %$_ : @$_;
    $_;
  } @rows;
}
```

Nice catch btw! Amazing that I didn't test/see this.

t/json.t
```diff
@@ -8,24 +8,40 @@ my $mysql = Mojo::mysql->new($ENV{TEST_ONLINE});
 my $db    = $mysql->db;
 
 eval {
-  $db->query('create table if not exists mojo_json_test (id int(10), name varchar(60), j json)');
```
jhthorsen, line 11: I don't think you can move this out of the eval block, since I want to skip the test if the mysql/mariadb version does not support JSON.

jhthorsen (changes requested):
I'm very positive to this change in general, just need a bit of tweaking.

lib/Mojo/mysql/Results.pm
Tekki: I think we have to, because after `fetchall` the statement handle is closed and we can no longer read from it.

lib/Mojo/mysql/Results.pm
Tekki: The first question is if we should a big `if ... else` or not. If we have 10000 rows we call `ref eq 'HASH' ?` for nothing 9999 times. There is a similar block below.
Then, without testing it, your proposal looks strange. We use `$_` in 2 different meanings and it seems you mix them. Notice that I added the `## no critic` which is already a bad sign.

t/json.t
Tekki: I've replaced it with a call to `json_type` because this is much faster than trying to create a table. My assumption was that a DB that has no json support doesn't understand this command. At least on MySQL 5.6 it works.

lib/Mojo/mysql/Results.pm
Tekki: > Then, without testing it, your proposal looks strange.

Don't trust the eyes, I should know that.  :see_no_evil:

t/json.t
Tekki: Works on MariaDB 10.1 too.

Tekki:
@jhthorsen What is missing here or what changes do I have to make? We should finish this before we start to look at #57.

lib/Mojo/mysql/Results.pm
jhthorsen: Then I suggest a bigger change:

```
# The method to call on $sth and true/false for deref
sub array { ($_[0]->_expand(fetchrow_arrayref => 0)[0] }
sub arrays { _c($_[0]->_expand(fetchall_arrayref => 1) }
```

I guess by doing that, we could merge _types() and _expand() again, and maybe even use that true/false flag instead of ref() inside _expand()?

jhthorsen:
> @jhthorsen What is missing here or what changes do I have to make? We should finish this before we start to look at #57.

Sorry for the late reply. I've made a comment now: https://github.com/jhthorsen/mojo-mysql/pull/58#discussion_r265103906

jhthorsen:
Here is my suggestion: (not tested)

```diff
diff --git a/lib/Mojo/mysql/Results.pm b/lib/Mojo/mysql/Results.pm
index cf4ca23..4a4e318 100644
--- a/lib/Mojo/mysql/Results.pm
+++ b/lib/Mojo/mysql/Results.pm
@@ -5,11 +5,13 @@ use Mojo::Collection;
 use Mojo::JSON 'from_json';
 use Mojo::Util 'tablify';

+my $LOOKS_LIKE_JSON_RE = qr/^[\[{].*[}\]]$/;
+
 has [qw(db sth)];

-sub array { ($_[0]->_expand($_[0]->sth->fetchrow_arrayref))[0] }
+sub array { ($_[0]->_expand(fetchrow_arrayref => 0, 0))[0] }

-sub arrays { _c($_[0]->_expand(@{$_[0]->sth->fetchall_arrayref})) }
+sub arrays { _c($_[0]->_expand(fetchall_arrayref => 0, 1)) }

 sub columns { shift->sth->{NAME} }

@@ -17,9 +19,9 @@ sub expand { $_[0]{expand} = defined $_[1] ? 2 : 1 and return $_[0] }

 sub finish { shift->sth->finish }

-sub hash { ($_[0]->_expand($_[0]->sth->fetchrow_hashref))[0] }
+sub hash { ($_[0]->_expand(fetchrow_hashref => 1, 0))[0] }

-sub hashes { _c($_[0]->_expand(@{$_[0]->sth->fetchall_arrayref({})})) }
+sub hashes { _c($_[0]->_expand(fetchall_arrayref => 1, 1, {})) }

 sub rows { shift->sth->rows }

@@ -42,34 +44,42 @@ sub state { shift->sth->state }
 sub _c { Mojo::Collection->new(@_) }

 sub _expand {
-  my ($self, @rows) = @_;
-
-  return @rows unless my $mode = $self->{expand} and $rows[0];
-
-  # Force expanding
-  return map {
-    my $r = $_;
-    $_ = from_json $_ for grep {/^(\[|\{).*(\}|\])/} values %$r;
-    $r;
-  } @rows if $mode == 2;
-
-  # Only expand json columns
-  my ($idx, $name) = @$self{qw(idx name)};
-  unless ($idx) {
-    my $types = $self->db->mysql->_dbi_attr($self->sth, 'type');
-    my @idx   = grep { $types->[$_] == 245 or $types->[$_] == 252 } 0 .. $#$types;    # 245 = MySQL, 252 = MariaDB
-    ($idx, $name) = @$self{qw(idx name)} = (\@idx, [@{$self->columns}[@idx]]);
-  }
-
-  return @rows unless @$idx;
-  if (ref $rows[0] eq 'HASH') {
-    for my $r (@rows) { $r->{$_} and ($r->{$_} = from_json $r->{$_}) for @$name }
+  my ($self, $sth_method, $is_hash, $is_list, @args) = @_;
+
+  my $mode = $self->{expand} || 0;
+  my ($idx, $names) = $mode == 1 ? $self->_types : ();    # Need to be done before getting $sql_data
+  my $sql_data = $self->sth->$sth_method(@args);
+  $sql_data = [$sql_data] unless $is_list;
+
+  if ($mode == 1) {
+    if ($is_hash) {
+      for my $r (@$sql_data) {
+        $r->{$_} = from_json $r->{$_} for grep { defined $r->{$_} } @$names;
+      }
+    }
+    else {
+      for my $r (@$sql_data) {
+        $r->[$_] = from_json $r->[$_] for grep { defined $r->[$_] } @$idx;
+      }
+    }
   }
-  else {
-    for my $r (@rows) { $r->[$_] and ($r->[$_] = from_json $r->[$_]) for @$idx }
+  elsif ($mode == 2) {
+    for my $r (@$sql_data) {
+      $_ = from_json $_ for grep $LOOKS_LIKE_JSON_RE, $is_hash ? values %$r : @$r;
+    }
   }

-  return @rows;
+  return @$sql_data;
+}
+
+sub _types {
+  my $self = shift;
+  return @$self{qw(idx names)} if $self->{idx};
+
+  my $types = $self->db->mysql->_dbi_attr($self->sth, 'type');
+  my @idx   = grep { $types->[$_] == 245 or $types->[$_] == 252 } 0 .. $#$types;    # 245 = MySQL, 252 = MariaDB
+
+  return ($self->{idx} = \@idx, $self->{names} = [@{$self->columns}[@idx]]);
 }
```

Tekki:
Looks quite similar to what I've seen before I thought jhthorsen doesn't like method names passed as string.

jhthorsen:
I don’t mind passsing method names. I do think those two numbers (ex: 0, 1) is kind of hard to understand (should probably be passed as %params instead), but I’ll make an exception, since it’s an internal method.

Please pick and choose what you want, but it does have some benefits:

- It doesn’t always call _types()
- It doesn’t have to call ref()

What I like the least is the nested if/else part when checking for mode==1 or 2, but I wasn’t able to golf it down.

Tekki:
I've made some changes. They are a bit different from what you proposed (and what I imagined), one reason is that we can't pass undefined arguments to the fetch methods.

Tekki:
> What I like the least is the nested if/else part when checking for mode==1 or 2, but I wasn’t able to golf it down.

What do you think of that?
```
  if ($mode == 1) {
    my $fn = $to->{hash}
      ? sub {
      $_[0]->{$_} = from_json $_[0]->{$_} for grep { defined $_[0]->{$_} } @$names;
      }
      : sub {
      $_[0]->[$_] = from_json $_[0]->[$_] for grep { defined $_[0]->[$_] } @$idx;
      };

    $fn->($_) for @$sql_data;
  }
```
Edit: We could extend this to include `$mode==2` too.

Tekki:
Something like
```
  if ($mode) {
    my %expand_fn = (
      1 => {
        array => sub {
          $_[0]->[$_] = from_json $_[0]->[$_] for grep { defined $_[0]->[$_] } @$idx;
        },
        hash => sub {
          $_[0]->{$_} = from_json $_[0]->{$_} for grep { defined $_[0]->{$_} } @$names;
        },
      },
      2 => {
        array => sub { $_ = from_json $_ for grep $LOOKS_LIKE_JSON_RE, @$_[0] },
        hash  => sub { $_ = from_json $_ for grep $LOOKS_LIKE_JSON_RE, values %{$_[0]} },
      },
    );

    my $do_expand = $expand_fn{$mode}{$to->{hash} ? 'hash' : 'array'};
    $do_expand->($_) for @$sql_data;
  }
```

jhthorsen:
Making a function makes it slower, so I don't think we should do that in this case.

Tekki:
For me it doesn't matter how we write it, but I find no significant speed difference between the two.
Instead I found that `grep $LOOKS_LIKE_JSON_RE` doesn't work, we have to write
`grep {$_ =~ $LOOKS_LIKE_JSON_RE}`
which makes the loop 15 % slower than
`grep /^[\[{].*[}\]]$/`

Tekki:
In any case the most expensive operation is `from_json`, followed by `grep`. Anything else than `grep /some-regex/` is extremely slow.

lib/Mojo/mysql/Results.pm
```diff
@@ -42,42 +44,53 @@ sub state { shift->sth->state }
...
-      $_ = from_json $_ for grep {$looks_like_json} ref eq 'HASH' ? values %$_ : @$_;
-      $_;
-    } @rows;
+  # Fetch sql data
+  my $sql_data;
+  if ($to->{list}) {
+    $sql_data = $to->{hash} ? $self->sth->fetchall_arrayref({}) : $self->sth->fetchall_arrayref;
+  }
+  else {
+    @$sql_data = $to->{hash} ? $self->sth->fetchrow_hashref : $self->sth->fetchrow_arrayref;
```
jhthorsen: Not a huge deal, but it's a bit easier for me to parse if it's written as `$sql_data = [...]`. Another way this could be written is:

```
my $sql_data
  = $to->{list} && $to->{hash} ? $self->sth->fetchall_arrayref({})
  : $to->{list} ? $self->sth->fetchall_arrayref
  : $to->{hash} ? [$self->sth->fetchrow_hashref]
  :               [$self->sth->fetchrow_arrayref];
```

...but I'm not sure if there's a huge win.

lib/Mojo/mysql/Results.pm
```diff
@@ -5,21 +5,23 @@ use Mojo::Collection;
 use Mojo::JSON 'from_json';
 use Mojo::Util 'tablify';
 
+my $LOOKS_LIKE_JSON_RE = qr/^[\[{].*[}\]]$/;
+
 has [qw(db sth)];
 
-sub array { ($_[0]->_expand($_[0]->sth->fetchrow_arrayref))[0] }
+sub array { ($_[0]->_expand({}))[0] }
```
jhthorsen: Just to make it easier for me to scan the code, could you change it to `{hash => 0, list => 0}`? ...same below, where only one argument is specified. Another idea is to write `{type => 'array'}` instead of `{hash => 0}`, since it will be easier to find the `$from_json` function that I suggest further down.

lib/Mojo/mysql/Results.pm
```diff
@@ -42,34 +42,53 @@ sub state { shift->sth->state }
...
+  if ($to->{list}) {
+    $sql_data = $to->{hash} ? $self->sth->fetchall_arrayref({}) : $self->sth->fetchall_arrayref;
   }
   else {
-    for my $r (@rows) { $r->[$_] and ($r->[$_] = from_json $r->[$_]) for @$idx }
+    @$sql_data = $to->{hash} ? $self->sth->fetchrow_hashref : $self->sth->fetchrow_arrayref;
+  }
+
+  # expand(), only expand json columns
+  if ($mode == 1) {
```
jhthorsen: Based on your comments on which part of the code that actually takes time to run, I suggest this:

```
sub _expand {
  ...
  my $from_json = __PACKAGE__->can(sprintf '_from_json_mode_%s_%s', $mode, $to->{hash} ? 'hash' : 'array');
  $from_json->($_, $idx, $names) for @$sql_data;
  ...
}

sub _from_json_mode_1_array {
  my ($r, $idx, $names) = @_;
  $r->[$_] = from_json $r->[$_] for grep { defined $r->[$_] } @$idx;
}

sub _from_json_mode_1_hash {
  my ($r, $idx, $names) = @_;
  $r->{$_} = from_json $r->{$_} for grep { defined $r->{$_} } @$names;
}

sub _from_json_mode_2_array {
  my ($r, $idx, $names) = @_;
  $_ = from_json $_ for grep /^[\[{].*[}\]]$/, @$r;
}

sub _from_json_mode_2_hash {
  my ($r, $idx, $names) = @_;
  $_ = from_json $_ for grep /^[\[{].*[}\]]$/, values %$r;
}
```

jhthorsen (commented):
More (potentially) nitpicking 😞

I'll be quicker at commenting/merging this change now, since I'm going to be working quite a bit in front of the computer the next days.

jhthorsen (approved):

jhthorsen:
LGTM 👍 Want to merge and make a release? Let me know if you want me to do it.

Tekki:
Don't trust your eyes, but it really looks better now.
Yes, I can merge it. We already make a new release or we wait for #57?

jhthorsen:
I don't mind many releases, so just go ahead and make a release if you like 👍 

Not sure if you want to update your contact before the release: https://github.com/jhthorsen/mojo-mysql/blob/bafbac9017157f83bc465d5611032db4bc29019b/lib/Mojo/mysql.pm#L541 (optional)

Tekki:
Yes, I will update it.
If we continue like that we will soon no longer be 'mostly a rip-off'.

PR #60, opened 2019-04-12
Optionally decode JSON

Tekki:
First commit adds failing tests, second a solution.

### DIFF
--- a/lib/Mojo/mysql/Results.pm
+++ b/lib/Mojo/mysql/Results.pm
@@ -67,12 +67,18 @@ sub _expand {
 
 sub _from_json_mode_1_array {
   my ($r, $idx, $names) = @_;
-  $r->[$_] = from_json $r->[$_] for grep { defined $r->[$_] } @$idx;
+  for (grep { defined $r->[$_] } @$idx) {
+    utf8::decode $r->[$_];
+    $r->[$_] = from_json $r->[$_];
+  }
 }
 
 sub _from_json_mode_1_hash {
   my ($r, $idx, $names) = @_;
-  $r->{$_} = from_json $r->{$_} for grep { defined $r->{$_} } @$names;
+  for (grep { defined $r->{$_} } @$names) {
+    utf8::decode $r->{$_};
+    $r->{$_} = from_json $r->{$_};
+  }
 }
 
 sub _from_json_mode_2_array {
@@ -90,7 +96,7 @@ sub _types {
   return @$self{qw(idx names)} if $self->{idx};
 
   my $types = $self->db->mysql->_dbi_attr($self->sth, 'type');
-  my @idx = grep { $types->[$_] == 245 or $types->[$_] == 252 } 0 .. $#$types;    # 245 = MySQL, 252 = MariaDB
+  my @idx   = grep { $types->[$_] == 245 or $types->[$_] == 252 } 0 .. $#$types;    # 245 = MySQL, 252 = MariaDB
 
   return ($self->{idx} = \@idx, $self->{names} = [@{$self->columns}[@idx]]);
 }
--- a/t/json.t
+++ b/t/json.t
@@ -7,20 +7,19 @@ plan skip_all => 'TEST_ONLINE=mysql://root@/test' unless $ENV{TEST_ONLINE};
 my $mysql = Mojo::mysql->new($ENV{TEST_ONLINE});
 my $db    = $mysql->db;
 
-eval {
-  $db->query('select json_type("[]")');
-} or do {
+eval { $db->query('select json_type("[]")'); } or do {
   plan skip_all => $@;
 };
 
+my $drink = 'szőlőlé';
 $db->query('create table if not exists mojo_json_test (id int(10), name varchar(60), j json)');
 $db->query('truncate table mojo_json_test');
-$db->query('insert into mojo_json_test (id, name, j) values (?, ?, ?)', $$, $0, {json => {foo => 42}});
+$db->query('insert into mojo_json_test (id, name, j) values (?, ?, ?)', $$, $0, {json => {drink => $drink}});
 
-is $db->query('select json_type(j) from mojo_json_test')->array->[0],             'OBJECT', 'json_type';
-is $db->query('select json_extract(j, "$.foo") from mojo_json_test')->array->[0], '42',     'json_extract';
-is_deeply $db->query('select id, name, j from mojo_json_test where json_extract(j, "$.foo") = 42')->expand->hash,
-  {id => $$, name => $0, j => {foo => 42}}, 'expand json';
+is $db->query('select json_type(j) from mojo_json_test')->array->[0],               'OBJECT',     'json_type';
+is $db->query('select json_extract(j, "$.drink") from mojo_json_test')->array->[0], qq|"$drink"|, 'json_extract';
+is_deeply $db->query(qq|select id, name, j from mojo_json_test where json_extract(j, '\$.drink') = '$drink'|)
+  ->expand->hash, {id => $$, name => $0, j => {drink => $drink}}, 'expand json';
 
 my $value_hash = {nick => 'supergirl'};
 my $value_json = Mojo::JSON::to_json($value_hash);


Grinnz:
This is incorrect. Correctly returned JSON will be a decoded string in which case utf8::decode will return the wrong value or fail.

Grinnz:
And you cannot guess if a value is UTF-8 encoded. The only possible workaround is to check for specifically known buggy server versions.

Tekki:
utf8::decode returns true or false and decodes the string in place.
If it's not correct what I make, you can probably show me a failing test. At the moment, t/json.t passes on all databases I've installed with both DBD::mysql and DBD::MariaDB.

fskale:
Are you sure that the system locales are UTF-8 aware  and you compiled it ?
What's the output of the command: <pre>locale</pre> ?
Should be something like:
<pre>
locale
LANG=en_US.UTF-8
LANGUAGE=en_US:en
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_PAPER="en_US.UTF-8"
LC_NAME="en_US.UTF-8"
LC_ADDRESS="en_US.UTF-8"
LC_TELEPHONE="en_US.UTF-8"
LC_MEASUREMENT="en_US.UTF-8"
LC_IDENTIFICATION="en_US.UTF-8"
LC_ALL=
</pre>
On debian:
reconfigure using: dpkg-reconfigure locales
On redhat based systems all locales are compiled and you only have to set the ENV accordingly.

Tekki:
> Are you sure that the system locales are UTF-8 aware and you compiled it ?

Yes of course, but how is this related to the topic we are discussing here?

pali:
> If it's not correct what I make, you can probably show me a failing test.

Your test just test identity. That you get same result as you inserted.

But it does not test if data are correctly written in database.

So if you have incorrect function for `json --> string encoding` and also incorrect/broken function for `string --> json encoding` then these tests does not have to catch it and could pass instead of failing.

To show that it is working you need test something like this:
```perl
ok $db->query('insert into mojo_josn_test (id, name, j) values (?, ?, ?)', $$, $0, qq|{"drink":"$drink"}|);
is_deeply $db->query('select json_extract(j, "$.drink") from mojo_json_test')->array->[0], qq|"$drink"|;
```
and
```perl
ok $db->query('insert into mojo_json_test (id, name, j) values (?, ?, ?)', $$, $0, {json => {drink => $drink}});
is_deeply $db->query('select j from mojo_josn_test')->array->[0], qq|{"drink":"$drink"}|);
```

Tekki:
Thanks @pali, I'll add more tests.

Tekki:
This one
```
is_deeply $db->query('select j from mojo_josn_test')->array->[0], qq|{"drink":"$drink"}|);
```
obviously doesn't work because this difference between MySQL and MariaDB is what this PR is about. At the moment we care about consistent results with `select(...)->expand`.

Grinnz:
You must know whether the result is encoded in order to decide to either utf8::decode or decode_json instead of from_json.

Grinnz:
For example try this: `my $double_encoded = encode 'UTF-8', $drink;` and insert that in a JSON. It may come back once or twice encoded depending whether the server is buggy, and if you always apply utf8::decode you will get the wrong result when it comes back correctly.

Tekki:
```
$db->query('insert into mojo_json_test (id, name, j) values (?, ?, ?)', $$, $0, encode('UTF-8', qq|{"drink":"$drink"}|));
```
and
```
is $db->query('select json_extract(j, "$.drink") from mojo_json_test')->array->[0], qq|"$drink"|, 'json_extract';
is_deeply $db->query(qq|select id, name, j from mojo_json_test where json_extract(j, '\$.drink') = '$drink'|)
  ->expand->hash, {id => $$, name => $0, j => {drink => $drink}}, 'expand json';
```
pass on both MySQL and MariaDB.
Btw. MariaDB is buggy because it treats JSON as string and not as binary object. :grin:

Grinnz:
Those are failures. The resulting value should be encoded to UTF-8, like the one you inserted.

Tekki:
No, the resulting value should always be Perl characters.
And look, I'm here to solve the problems that @ViktorNacht reported and not to kill the time until the start of the Shanghai GP.

Grinnz:
They are Perl characters. The result of encode('UTF-8', $str) is the bytes representing $str in UTF-8 format, but it is also a string of characters that can be encoded to UTF-8 itself to be represented. This is why you cannot distinguish characters and bytes in Perl strings programatically.

The string you inserted is encode('UTF-8', $drink), not $drink. It must round trip. I'm sorry that you think I am wasting time, but I cannot abide grievous encoding errors to be added to another module.

Tekki:
We change the `->expand` method that automatically expands JSON fields and nothing more. If it doesn't fit your needs, you can either process the raw data directly or show a real use case where it fails. If you do this, the method will be changed again. But until then I think the `ut8::decode` way works, even if it doesn't look beautiful from an aesthetical perspective.

Grinnz:
It has nothing to do with aesthetics. It is wrong behavior. Encoding is not a guessing game. I already gave a case where this guessing fails.

Grinnz:
Like I said at the beginning there is only one correct way to workaround this. Figure out what combinations of mysql and DBD::mysql are buggy, and in only those cases, use decode_json instead of from_json.

Tekki:
So it's not about beautiful/ugly, but about correct/wrong. I prefer to look at it from an evolutionary point of view and let the 'guessing game' live until the selection pressure gets too high.

Tekki:
The Chinese GP wasn't so interesting, so I had some more time to look into this.
My 'guessing game' deals with the different data types returned by MySQL and MariaDB. I don't think it introduces any problems and it would make a difference if we duplicate the methods into one that uses `decode_json` and another one that uses `from_json`.
But now there is something else. Pali mentioned that we have to check how the data is inserted into the database. If we take your example with `$double_encoded = encode 'UTF-8', $drink` and send these UTF-8 bytes to the db, DBD::mysql and DBD::MariaDB don't seem to to the same. The first interprets the string as 'already encoded', the second as 'unencoded'. This means the query
```
select json_extract(j, "$.drink") from mojo_json_test
```
will return `$drink` with DBD::mysql and `$double_encoded` with DBD::MariaDB. At this place, we can't do anything about it, besides probably to add a warning to the README.

Grinnz:
That is because of the well known broken behavior of DBD::mysql with encoding which necessitated the DBD::MariaDB fork to begin with. You can apply `utf8::upgrade $double_encoded` to force DBD::mysql to use it as intended.

Grinnz:
You can read more about such a workaround in this documentation which was never released for DBD::mysql since the 4.042 change was reverted. https://github.com/perl5-dbi/DBD-mysql/pull/119/files

Tekki:
I remember you have told us this before. Just me personally, I don't remember I ever had encoding issues with DBD::mysql. And I rather think is another discussion because I don't see how this is related to the changes I've made to `Result->expand`.

Grinnz:
It is the cause of the discrepancy you mentioned. It's not a new problem from these changes.

pali:
Look, DBD::mysql has seriously broken Unicode support and DBD::MariaDB was created just to fix these problems. @Grinnz already mentioned some hacks/workarounds for it, but why trying to support broken/double-encoded/double-decoded and similarly corrupted non-ASCII data with DBD::mysql and bring these problems back, if they are already fixed in DBD::MariaDB? These problems cannot be fixed in DBD::mysql due to legacy reasons, but why to create more **legacy**/broken application?

Tekki:
The changes in this PR have nothing to do with the UTF-8 handling in DBD::mysql.

jhthorsen:
Could this be set as an option on the Mojo::mysql object? That way, I don't think we need to discuss it anymore, if it's something you need to opt in to get. Example:

```
$mysql->options->{mojo_mysql_utf8_decode_json} = 1;

# Later in Mojo::mysql::Result:
if ($self->db->mysql->options->{mojo_mysql_utf8_decode}) utf8::decode $r->[$_];
```

Tekki:
I don't think this should be an option. The target is that
```
$db->query('select json_extract(j, "$.drink") from mojo_json_test')->array->[0]
```
and
```
$db->query(qq|select id, name, j from mojo_json_test|)->expand->hash->{drink}
```
return the same value *under any circumstances*. Most of the discussion in this thread was about other things.

pali:
> return the same value *under any circumstances*

Exactly! When it does not return same value then you are going to corrupt data for application which called one of those function.

Grinnz:
> Most of the discussion in this thread was about other things.

It wasn't. The only discussion about other things is the UTF-8 encoding bug in DBD::mysql, which was referenced because your tests need to work around it. All of the other discussion is relevant to this issue. I have specified the only correct workaround twice, arbitrary wrong decoding will never be correct.

Tekki:
`t/json.t` produces exactly the same errors with DBD::mysql and DBD::MariaDB. If it's wrong how I solve them, there must be a test that fails because of this change.

jhthorsen:
I'm closing this, since it seems to have gone stale.

Issue #59, opened 2019-03-12
Abstract USING keyword in JOIN syntax


ViktorNacht:
I'm one of those weird people that adds the singular table name to the id column (e.g. foo_id), which is especially nice for simplifying JOINS:

`SELECT * FROM foo, bar USING (foo_id)`

This seems verbose:

`$db->select(['foo', ['bar', 'foo.foo_id' => 'bar.foo_id' ]);`

(which is what I'm using in my current app)

For the sake of maximum flexibility and completeness, I just wanted to wishlist the option of doing something like this:

```
$db->select(['foo', ['bar', 'foo_id']);
$db->select(['foo', ['bar', [qw(foo_id foo_col2 foo_col3)]]);
```

Just a suggestion. Feel to close this.

jhthorsen:
How can you tell the difference between a table ad column name?

Tekki:
The correct query seems to be
```
SELECT * FROM foo JOIN bar USING (foo_id)
```
For the sake of completeness we could implement the `NATURAL JOIN` too in something like
```
$db->select(['foo', [-natural => 'bar']]);
```
The question is always, where does it make more sense to write a query directly with `$db->query()`? If we implement everything that is imaginable we end up with a module that is unusable. Remember HTML generation with CGI?
By the way, I'd prefer to see `-natural` supported, as the syntax would be consistent with what we already use for left and right joins.

jhthorsen:
@Tekki: Feel free to implement or request a PR 😄

ViktorNacht:
> For the sake of completeness we could implement the `NATURAL JOIN` too in something like
> 
> ```
> $db->select(['foo', [-natural => 'bar']]);
> ```
> By the way, I'd prefer to see `-natural` supported, as the syntax would be consistent with what we already use for left and right joins.

This clearly makes way more sense given the current structure and is very clear and succinct.

My argument against falling back on `$db->query()` is it becomes difficult if not impossible to build SQL across multiple helpers and/or using dynamic controllers, etc. I know that's stating the obvious, but I think this module is wonderful alternative to DBIC.

Tekki:
I could imagine something like that:
```
# -natural
# "select * from foo natural join bar"
$abstract->select(['foo', [-natural => 'bar']]);

# using
# "select * from foo join bar using (foo_id)"
$abstract->select(['foo', [bar => 'foo_id']]);
```

Tekki:
> I could imagine something like that:
> 
[...]

PR #61, opened 2019-04-19
Add support for NATURAL JOIN and JOIN USING

Tekki:
These are the changes for #59.

### DIFF
--- a/lib/SQL/Abstract/mysql.pm
+++ b/lib/SQL/Abstract/mysql.pm
@@ -239,7 +239,7 @@ names, but also array references with tables to generate C<JOIN> clauses for.
   $abstract->select(['foo', ['bar', foo_id => 'id']]);
 
   # "select * from foo join bar on (foo.id = bar.foo_id)"
-  $abstract->select(['foo', ['bar', foo.id => 'bar.foo_id']]);
+  $abstract->select(['foo', ['bar', 'foo.id' => 'bar.foo_id']]);
 
   # -left, -right, -inner
   # "select * from foo left join bar on (bar.foo_id = foo.id)"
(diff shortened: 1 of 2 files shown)


lib/SQL/Abstract/mysql.pm
```diff
@@ -228,12 +239,20 @@ names, but also array references with tables to generate C<JOIN> clauses for.
   $abstract->select(['foo', ['bar', foo_id => 'id']]);
 
   # "select * from foo join bar on (foo.id = bar.foo_id)"
-  $abstract->select(['foo', ['bar', 'foo.id' => 'bar.foo_id']]);
+  $abstract->select(['foo', ['bar', foo.id => 'bar.foo_id']]);
```
jhthorsen: I think you should keep the quotes :)

jhthorsen (approved):
I can’t really find any issues with this, so feel free to merge it when you want 👍

The only thing I was wondering is if we should add experimental support for just one argument, a table name, then automatically use table_id or id_table as key... but maybe it won’t be possible to come to an agreement which one is the preferred one? 😉

Btw: That comment above is out of scope, so it shouldn’t affect the time to merge.

lib/SQL/Abstract/mysql.pm
Tekki: I thought TMTOWTDIBSCINABTE 😀 
Edit: But of course it's not consistent.

Tekki:
> The only thing I was wondering is if we should add experimental support for just one argument, a table name, then automatically use table_id or id_table as key... but maybe it won’t be possible to come to an agreement which one is the preferred one? 😉

Or interpret one argument as NATURAL JOIN? We can try such things, but only if all discussions with graduates in Defense Against the Dark Arts are automatically assigned to you.

jhthorsen:
Btw: You can make a new release if you like.

PR #62, opened 2019-04-23
Added missing prereq DBI as suggested by CPANTS.

manwar:
Hi @jhthorsen 

Please review the PR.
https://cpants.cpanauthors.org/release/TEKKI/Mojo-mysql-1.15

Many Thanks.
Best Regards,
Mohammad S Anwar

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -1,5 +1,6 @@
 requires 'Mojolicious'   => '8.03';
 requires 'DBD::mysql'    => '4.042';
 requires 'SQL::Abstract' => '1.86';
+requires 'DBI'           => '1.627';
 
 test_requires 'Test::More' => '0.90';


jhthorsen (changes requested):
Thanks for the PR 👍 

I don't understand how DBD::mysql does not depend on DBI.

Anyways, I think if we depend on DBI, we should have a minimum version number. How about 1.609, since that's what DBD::mysql seems to have in Makefile.PL and it's later than what DBD::MariaDB require. Or 1.627, since it's available in Centos7. Both are quite old, but I haven't read the Changelog thoroughly enough to say if we need any more recent.

Tekki:
I think this test produces some false alerts. For [Mojolicious](https://cpants.cpanauthors.org/release/SRI/Mojolicious-8.14) for example it complains about EV and Sub::Util.

jhthorsen:
@Tekki: Thanks for the insight! I don't mind adding DBI as a prereq though, since it is something that is needed for the Mojo::mysql to work.

Grinnz:
It is because DBI is explicitly used (https://github.com/jhthorsen/mojo-mysql/blob/master/lib/Mojo/mysql.pm#L5) and the test does not know if you can rely on the DBD::mysql dependency to provide this (even though we know logically it must). Also it is nice to have the separate dependency in case Mojo::mysql needs a specific version of DBI that DBD::mysql doesn't.

jhthorsen:
So what does that conclude on? Please add thumbs up/down on this:

```
requires 'DBI' => 1.627;
```

Tekki:
What speaks against adding DBD::MariaDB too?
```
requires 'DBI' => 1.627;
requires 'DBD::MariaDB' => 1.21;
requires 'DBD::mysql' => '4.042';
```

jhthorsen:
> What speaks against adding DBD::MariaDB too?

I don't think we'll ever get done with this PR if we start discussing that as well. Please open an issue where you suggest that, or replacing DBD::mysql with DBD::MariaDB as default, or something else. I really don't want to mix these two discussions here.

Grinnz:
> ```
> requires 'DBI' => 1.627;
> ```

Quoted like the other versions :)

jhthorsen:
@manwar: Care to update the PR with the suggested version number?

manwar:
Updated as suggested.
Please review again when you get time.

jhthorsen (approved):

jhthorsen:
Thank you! 👍

Issue #75, opened 2020-04-28
SQL::Abstract::mysql overrides random parts of the SQL::Abstract internals and will almost certainly break with the next SQL::Abstract release


shadowcat-mst:
If you want additional features in SQL::Abstract please contact the SQL::Abstract development team via irc.perl.org #dbix-class - ping 'mst' or 'ilmari' specifically.

The SQL::Abstract::mysql shipped as part of Mojo::mysql is incredibly fragile and since nobody bothered talking to us before writing it, relies on things that are unstable, unsupported and will almost certainly be broken by the next release of SQL::Abstract.

The current code should never have been released to cpan as a stable release and I'm really quite sad that, unlike sri (the author of Mojo::Pg), it didn't occur to whoever perpetrated this that co-operating with the developers of other cpan modules is a good idea, especially if you're going to override undocumented and unsupported parts of the internals and then have that talk to other people's production databases.

-- mst

jhthorsen:
@Tekki: Is this something you have time to fix in a stable way? If not, then I think I have to mark SQL::Abstract::mysql as DEPRECATED, and make a release with a huge warning. I find the SQLA code very difficult to navigate and override, so I don't see how I can find time to fix it myself.

I will however have a second look if nobody else can.

To every user of Mojo::mysql: I'm very sorry that I accepted changes that would get delete(), insert() and select() into the state they are now. query() is not affected though, so if you craft your own SQL, then you have nothing to worry about.

Note to self: Do *not* accept pull requests where the code override private methods and/or methods marked as experimental, without talking to the maintainers.

shadowcat-mst:
OK. I believe I have *mostly* figured this out. I have no idea at all what the _METHOD_for_refkind crap was in aid of but here's how you do it without that:

    diff --git a/lib/SQL/Abstract/mysql.pm b/lib/SQL/Abstract/mysql.pm
    index f97a1a2..39927dc 100644
    --- a/lib/SQL/Abstract/mysql.pm
    +++ b/lib/SQL/Abstract/mysql.pm
    @@ -5,12 +5,11 @@ BEGIN { *puke = \&SQL::Abstract::puke }
     
     sub insert {
       my $self    = shift;
    -  my $table   = $self->_table(shift);
    -  my $data    = shift || return;
    -  my $options = shift || {};
    +  my $table   = $self->_table($_[0]);
    +  my $data    = $_[1] || return;
    +  my $options = $_[2] || {};
     
[...]

PR #76, opened 2020-04-30
Depending less on the internals of SQL::Abstract #75

jhthorsen:
This PR tries to address the issues addressed in #75. It's still using some internal methods, but I hope this is a step in the right direction.

Methods used from SQL::Abstract are:
* _quote()
* _recurse_where()
* _sqlcase()
* _SWITCH_refkind()
* _update_set_values()

Please let me know if it's still a "train wreck" @shadowcat-mst or if there's some modifications I can do to use fewer internal methods.


jhthorsen:
@shadowcat-mst did not approve.

shadowcat-mst:
The t/sql.t changes were perfect.

The lib/SQL/Abstract/mysql.pm changes need only to be what I already provided in the ticket.

PR #78, opened 2020-04-30
Allow where() with complex options to be called by itself

jhthorsen:
Split out the `where()` change from #76 into a separate PR.

This change allows users to call `$abstract->where()` without having to go through `select()`.

This change also split out the sql creating into "_mysql_xxx" methods, since I find that a lot easier to read. This means however that the diff might be a bit bigger than it should be, so maybe it's easier to look at the final result.


jhthorsen:
@shadowcat-mst went to sleep, so I decided to merge and make a release. I'll take his feedback in the morning (if any) and make another release if I have to.

PR #80, opened 2020-06-12
Make Mojo::mysql::PubSub less susceptible to deadlocks/timeouts

larryl:
We ran into problems with deadlocks and timeouts when running 4+ Minion workers against the same database.  After that kind of failure, the worker would continue running but the pub/sub logic would spin the CPU up to 100% and stay there.  This pull request adds simple one-time retries of deadlock/timeout failures to pub/sub queries.  We've been running this fix for over a month and have not see any problems since making the fix.


lib/Mojo/mysql/PubSub.pm
```diff
@@ -100,14 +131,57 @@ sub _wait_db {
...
+    $wait_db,
+    'select sleep(600)',
+    $cb
+  );
+}
+
+sub _query_with_retry {
+  my ($db, @query_args) = @_;
+
+  my $wantarray = wantarray;
```
jhthorsen: Why is `wantarray` added? I would appreciate if it was removed.

lib/Mojo/mysql/PubSub.pm
```diff
@@ -100,14 +131,57 @@ sub _wait_db {
...
+  my $remaining_attempts = 2;
+  while ($remaining_attempts--) {
+    local $@;
+    eval {
+      if    (!defined $wantarray) {               $db->query(@query_args) }
+      elsif          ($wantarray) { @results    = $db->query(@query_args) }
+      else                        { $results[0] = $db->query(@query_args) }
+    };
+    last unless $@;  # success
+    die unless $remaining_attempts;  # rethrow $@ if no remaining attempts
```
jhthorsen: Could you change `die` to `croak $@` instead? (Same below)

lib/Mojo/mysql/PubSub.pm
```diff
@@ -100,14 +131,57 @@ sub _wait_db {
...
+
+    # If we are allowed to retry, check if the error message looks
+    # like it refers to something retryable.  Only look within the
+    # first line to avoid potential spurious matches if the error
+    # e.g. contains a stack trace.
+    my $err = $@;                                  # avoid stringifying $@ ...
+    die unless $err =~ /^\V*(?:retry|timeout)/i;   # ... and maybe rethrow it
+
+    # If we got here, we are retrying the query:
+    warn qq|[PubSub] (@{[$db->pid]}) Retrying after: !!! $err\n| if DEBUG;
```
jhthorsen: Not sure how useful this is: Does it say anything about what it will retry? How about changing to something like:

```
my ($self, $sql, @bind) = @_; # instead of my ($db, @query_args) = @_;
...
warn qq|[PubSub] (@{[$db->pid]}) retry ($sql) !!! $err\n|
```

lib/Mojo/mysql/PubSub.pm
```diff
@@ -238,7 +312,7 @@ drop table if exists mojo_pubsub_subscribe;
 drop table if exists mojo_pubsub_notify;
 
 create table mojo_pubsub_subscribe (
-  id integer auto_increment primary key,
+  id bigint unsigned auto_increment primary key,
```
jhthorsen: You need to add a new `-- 2 down` and `-- 2 up` to make schema changes. (Same below)

lib/Mojo/mysql/PubSub.pm
```diff
@@ -100,14 +131,57 @@ sub _wait_db {
...
+  );
+}
+
+sub _query_with_retry {
+  my ($db, @query_args) = @_;
+
+  my $wantarray = wantarray;
+  my @results;
+
+  my $remaining_attempts = 2;
```
jhthorsen: Should this be configurable? Maybe add an attribute or environment variable? Might be useful for someone to set it to 0? Example:

```
my $remaining_attempts = $self->number_of_retries + 1;
```

...though I'm not sure if `number_of_retries()` is a good attribute name.

jhthorsen (changes requested):
Sorry for the late reply. I'm unsure when this happens, but I'll take the PR when the changes have been done.

Also: I don't think the code is run through `perltidy` with the `.perltidyrc` from this repo.

lib/Mojo/mysql/PubSub.pm
larryl: Will do - I'm used to working with DBIx::Class where it can return different stuff depending on calling context, but looking at `Mojo::mysql::Database::query` I see where it only ever returns a single scalar.

lib/Mojo/mysql/PubSub.pm
larryl: Good idea, I will try making it an attribute.  Maybe `max_allowed_retries` or `max_allowed_query_retries`?

lib/Mojo/mysql/PubSub.pm
larryl: I think the error includes the SQL but I can't find a copy of the error text right now.  Probably can't hurt to explicitly include the SQL in the warning though.

lib/Mojo/mysql/PubSub.pm
jhthorsen: I think max_allowed_retries() or just adding an environment variable might be an alternative as well:

use constant RETRIES => $ENV{MOJO_MYSQL_PUBSUB_RETRIES} || 2;

lib/Mojo/mysql/PubSub.pm
larryl: I think I'm just going to pull this change.  Without using DDL tools, there's no safe way to change the column def on a live database without locking the table.  Plus it was more an issue before I realized the `REPLACE` commands were chewing up primary keys, and the `INSERT ... ON DUPLICATE KEY UPDATE` gets rid of that problem.

lib/Mojo/mysql/PubSub.pm
larryl: Fixed in b46789c

lib/Mojo/mysql/PubSub.pm
larryl: Fixed in 62fe1c3

lib/Mojo/mysql/PubSub.pm
larryl: Fixed in 9cf53de

lib/Mojo/mysql/PubSub.pm
larryl: Reverted in 3bbdbe6

lib/Mojo/mysql/PubSub.pm
larryl: Fixed in c9d2f95

larryl:
All the recommendations have been addressed, also fixed the perltidying in 60f0055

jhthorsen:
Thanks for making the requested changes 👍

PR #86, opened 2021-10-05
Fix last_insert_id and other sth_attrs in Results

Grinnz:
This line seems to be for the purpose of setting a value to the attribute, but this is never done with _sth_attr (only _dbh_attr) and is broken because $self and $name still remain in `@_` so this line will always run, attempting to set the attr to undef. This causes an error in the case of readonly attributes like mariadb_insertid, or would cause issues in other code where attr should not be set to undef.


jhthorsen:
Thanks!

PR #92, opened 2023-04-06
PubSub retry also on deadlock

andrii-suse:
1. Always log failed error message in retry when PUBSUB_DEBUG is enabled.

After that I started seeing in debug output:
```
[PubSub] (6093845) noretry (insert into mojo_pubsub_subscribe (pid, channel) values (?, ?) on duplicate key update ts=current_timestamp) !!! DBD::mysql::st execute failed: Deadlock found when trying to get lock; try restarting transaction
```

2. Retry when DB reports a deadlock.

See also report about deadlock in https://jira.mariadb.org/browse/MDEV-31017

### DIFF
--- a/lib/Mojo/mysql/PubSub.pm
+++ b/lib/Mojo/mysql/PubSub.pm
@@ -137,10 +137,13 @@ sub _query_with_retry {
     # first line to avoid potential spurious matches if the error
     # e.g. contains a stack trace.
     my $err = $@;                                        # avoid stringifying $@ ...
-    croak $@ unless $err =~ /^\V*(?:retry|timeout)/i;    # ... and maybe rethrow it
-
-    # If we got here, we are retrying the query:
-    warn qq|[PubSub] (@{[$db->pid]}) retry ($sql) !!! $err\n| if DEBUG;
+    if ($err =~ /^\V*(?:retry|timeout|try restart)/i) {  # ... and maybe rethrow it
+      # If we got here, we are retrying the query:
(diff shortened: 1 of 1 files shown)


lib/Mojo/mysql/PubSub.pm
```diff
@@ -137,10 +137,13 @@ sub _query_with_retry {
...
     # e.g. contains a stack trace.
     my $err = $@;                                        # avoid stringifying $@ ...
-    croak $@ unless $err =~ /^\V*(?:retry|timeout)/i;    # ... and maybe rethrow it
-
-    # If we got here, we are retrying the query:
-    warn qq|[PubSub] (@{[$db->pid]}) retry ($sql) !!! $err\n| if DEBUG;
+    if ($err =~ /^\V*(?:retry|timeout|try restart)/i) {  # ... and maybe rethrow it
+      # If we got here, we are retrying the query:
+      warn qq|[PubSub] (@{[$db->pid]}) retry ($sql) !!! $err\n| if DEBUG;
+    } else {
```
jhthorsen, line 143: This PR does not conform with the codestyle. Please run perltidy and force push the changes.

jhthorsen (changes requested):
This change has no tests so there's no guarantee that the fix won't get reverted at a later point in time. Please add tests if you want to prevent that from happening.

I'll take the change regardless though if you want me to, since the ::PubSub class is not really a supported part of Mojo::mysql.

